PR #579, opened 2019-09-09
use babel-plugin-transform-async-to-promises

### LINKED ISSUE #23: Implement fast-async in the ft-js plugin to help keep the size of the generator output for async/await reasonable
https://github.com/MatAtBread/fast-async

apaleslimghost:
fixes #23

### DIFF
--- a/examples/kitchen-sink/client/main.js
+++ b/examples/kitchen-sink/client/main.js
@@ -5,7 +5,8 @@ import * as appContext from '@financial-times/dotcom-ui-app-context'
 import * as tracking from '@financial-times/n-tracking'
 import * as ads from '@financial-times/n-ads'
 
-domLoaded.then(() => {
+async function main() {
+  await domLoaded
   const flagsClient = flags.init()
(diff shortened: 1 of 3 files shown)


nickcolley:
I just ran your reduced test case and it looks like the newer version outputs correctly:

```js

var getStatus = function getStatus() {
  try {
    var response;
    var url = "/path/to/status";
    return Promise.resolve(_catch(function () {
      return Promise.resolve(fetch(url)).then(function (_fetch) {
        response = _fetch;

        if (response.ok) {
          return response.json();
        } else {
          throw Error("".concat(url, " responded with ").concat(response.status));
        }
      });
    }, function (error) {
      console.error(error);
    }));
  } catch (e) {
    return Promise.reject(e);
  }
};
```

Does this mean this can continue and is no longer blocked?

apaleslimghost:
closing this as we've clearly been fine without it for three years and also i'm pretty sure `async`/`await` is supported in all our target browsers now

PR #646, opened 2019-10-31
don't output wrapper div if contents is a react node

apaleslimghost:
because `contents` is either a string or a `ReactNode`, we can be sure it's a `ReactNode` here and return it directly (having said that, it doesn't typecheck 🤔)


apaleslimghost:
i can't approve this since i opened it originally but Matt's changes look great :)

PR #719, opened 2020-01-13
Adds syndicatable subnav item ids to the myFT subnav component

pixelandpage:
Adds a data-id attribute to subnav list items on the myFT pages. 

The attribute is present in [n-ui]( https://github.com/Financial-Times/n-ui/blob/master/components/n-ui/header/partials/subnav/template.html#L24) and is required by n-syndication.

### DIFF
--- a/packages/dotcom-ui-header/src/components/sub-navigation/partials.tsx
+++ b/packages/dotcom-ui-header/src/components/sub-navigation/partials.tsx
@@ -73,9 +73,10 @@ const SubSections = ({ items }: { items: TNavMenuItem[] }) => {
       data-trackable="subsections">
       {items.map((item, index) => {
         const selected = item.selected ? 'o-header__subnav-link--highlight' : ''
+        const subnavItemId = { 'data-id': item.id ? `subnav-${item.id}` : null }
 
         return (
-          <li className="o-header__subnav-item" key={`item-${index}`}>
+          <li className="o-header__subnav-item" key={`item-${index}`} {...subnavItemId}>
             <a
               className={`o-header__subnav-link ${selected}`}
               href={item.url}
--- a/packages/dotcom-types-navigation/index.d.ts
+++ b/packages/dotcom-types-navigation/index.d.ts
@@ -47,6 +47,7 @@ export type TNavMenuItem = {
   label: string
   url: string | null
   submenu?: TNavMenu | TNavMenuWithColumns
+  id?: string
   selected?: boolean
   meganav?: TNavMeganav[]
   disableTracking?: boolean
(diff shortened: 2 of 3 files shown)


packages/dotcom-ui-header/src/components/sub-navigation/partials.tsx
```diff
@@ -73,9 +73,10 @@ const SubSections = ({ items }: { items: TNavMenuItem[] }) => {
       data-trackable="subsections">
       {items.map((item, index) => {
         const selected = item.selected ? 'o-header__subnav-link--highlight' : ''
+        const subnavItemId = { 'data-id': item.id ? `subnav-${item.id}` : null }
```
andygout: It is possible to add the `data-id` property to the object only if the value upon which it depends exists, rather than adding it by default (and setting its value to `null` when the dependent value does not exist).

```
const listItemProps = {
   ...(item.id && { 'data-id': `subnav-${item.id}` })
}
```

And then:
 ``<li className="o-header__subnav-item" key={`item-${index}`} { ...listItemProps }>``

This will also mean we do not need to add these instances of `data-id={null}` to `packages/dotcom-ui-header/src/__test__/components/__snapshots__/MainHeader.spec.tsx.snap`.

It might be worth adding a test to ensure that the correct `data-id` value is output when an `item.id` value is provided (bearing in mind the opinions devs shared about snapshot tests in the last dev huddle).

pixelandpage: I had this same concern this morning but it looks like React won't output the attribute if it's set to `null`, (jest builds the components differently) so we should be safe with this. We've used the same pattern in a few other places in Page Kit so it's internally consistent.

andygout: Okay, so it's a toss-up between:

- Write code that is built by our build process in line with how Jest builds, meaning that what we test in Jest is more reliably a reflection of what we build in production.
- Keep the pattern consistent throughout Page Kit.

I think there is a strong case for both so will leave the decision to you.

i-like-robots:
> The attribute is present in n-ui and is required by n-syndication.

Do we know why n-syndication needs it?

pixelandpage:
> Do we know why n-syndication needs it?

So, I had a strong suspicion, which I've now basically confirmed, that this behaviour is an accident. 

Because the markup in Page Kit differs from n-ui I started seeing `POST https://local.ft.com:5050/syndication/resolve 400 (Bad Request)` errors in the console on the Contact Preferences page, which doesn't actually contain any elements which we would expect to be working with syndication. 

On the live site, requests to n-syndication on myFT pages are accidentally including the subnav item ids. This means that the request to `/syndication/resolve` is not empty. In my branch, these ids were not being picked up by n-syndication (the correct behaviour), but the empty response payload coming back from the service triggered the errors. 

I'll create an issue in n-syndication to fix this properly 🙂

PR #720, opened 2020-01-13
Remove 'top' as a oHeader sass option

pixelandpage:
Removes 'top' as a sass option in dotcom-ui-header as it's no longer an option on oHeader https://github.com/Financial-Times/o-header#sass, This doesn't seem to affect prod but was causing errors when building storybook locally.


i-like-robots:
Probably need to rebase on #721 when that's merged, a 👍 for when it goes green

PR #723, opened 2020-01-14
Bump n-ads

alexflorisca:
Bumped n-ads to latest version. n-ads v2 updates all the dependencies to latest version. n-ads v3 now requires `appContext` to be passed in as an option when being initialised.


pixelandpage:
As this bumps the package by a couple of major versions would we be able to get some context about the change and what will be impacted in the PR description?

PR #732, opened 2020-01-22
add live-blog as a valid content type

apaleslimghost:
one more from article smoke test whack-a-mole


apaleslimghost:
@i-like-robots do i need to regenerate the docs?

i-like-robots:
Ooh, yes - please do. There is an npm script in the app context package to do so.

i-like-robots:
cc https://github.com/Financial-Times/next-es-interface/issues/1098

PR #741, opened 2020-01-31
Add showLogoLink option to logo only header

remybach:
Please let me know if I've missed anything but as far as I can tell, this option will be able to be passed through using the `headerOptions` of the `Layout` component.

### DIFF
--- a/packages/dotcom-ui-header/README.md
+++ b/packages/dotcom-ui-header/README.md
@@ -26,7 +26,7 @@ This package provides several UI components to render different parts and styles
 
 - `<Header />` the full header with navigation with lots of options. See [header elements](#header-elements) for a breakdown of its parts.
 - `<Drawer />` the navigation drawer which should be rendered separately from the header, preferably near the bottom of the document.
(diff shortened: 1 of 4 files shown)


remybach:
@andygout: That was always the default way we used this logo only header before so I think we may want to keep it backwards compatible.

packages/dotcom-ui-header/README.md
```diff
@@ -26,7 +26,7 @@ This package provides several UI components to render different parts and styles
 
 - `<Header />` the full header with navigation with lots of options. See [header elements](#header-elements) for a breakdown of its parts.
 - `<Drawer />` the navigation drawer which should be rendered separately from the header, preferably near the bottom of the document.
-- `<LogoOnly>` a simple masthead displaying the logo image which does require any configuration.
+- `<LogoOnly />` a simple masthead displaying the logo image which does require any configuration. Doesn't link to the homepage by default, pass in a `showLogoLink={true}` prop to make it link through.
```
i-like-robots: > which does require any configuration

🤔 😄

i-like-robots:
I've switched the PR base to the `development` branch just in case we need to release any bug fixes from `master`.

Issue #740, opened 2020-01-30
Replace `dom-loaded` with `ready-state`


i-like-robots:
Replace https://www.npmjs.com/package/dom-loaded with https://www.npmjs.com/package/ready-state

We've noticed the requirement for several apps and components to wait until the document ready state is complete or document load event is triggered. The `ready-state` package includes promises for DOM loaded and interactive and loaded ready states.

This will require changes to the example apps, migration guide, and code splitting configuration.

PR #742, opened 2020-02-03
Replace dom-loaded with ready-state

andygout:
Fixes: https://github.com/Financial-Times/dotcom-page-kit/issues/740.

Once this PR is approved I will make the necessary changes in the migration docs: https://github.com/Financial-Times/dotcom-page-kit/wiki/Migrating-from-n%E2%80%93ui-to-Page-Kit.

@apaleslimghost Re. your [comment](https://github.com/Financial-Times/n-tracking/pull/39/files#r373430152) in this n-tracking PR:
- Should we also be using the `ready-state` package in that repo (or have Page Kit wrap it in a function that it can use)?
- Do we also need to identify which other apps that consume Page Kit likewise need this change applied?

#### New dependencies:
- [npm: ready-state](https://www.npmjs.com/package/ready-state).

### DIFF
--- a/docs/design-decisions/app-context.md
+++ b/docs/design-decisions/app-context.md
@@ -1,0 +1,17 @@
+# Design Decisions: App context
+
+App context is a combination of metadata about the current application and the page it is serving for use by client-side code. This is a change from the existing FT.com toolset.
+
+Previously metadata was embedded into the HTML as data attributes on the document element (`<html>`). This implementation had several problems:
+
+- What the metadata properties were and how they should be used was not defined anywhere.
+- The mechanism was often misused to pass additional information from server to client which lead to de facto standards emerging across different parts of FT.com.
+- We did not know who or what relies on this data meaning we were scared to change it, or fix it.
+- Several properties had changed in scope, either the name became irrelevant or the type of data had changed.
+
+To help resolve this our implementation of app context is based upon a schema which documents and guarantees the data. The schema also defines how the feature can be used to reduce the chance of it being re-purposed or misused. Because we don't always know who or what may be using the data the schema enables us communicate an official point of reference for them to follow.
+
+The new app context feature has also enabled us to refactor the [ads] and [tracking] component configuration to be deterministic rather than have each separately and arbitrarily scraping pages for information.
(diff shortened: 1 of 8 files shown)


i-like-robots:
- https://github.com/Financial-Times/next-subscribe/blob/496e35570989f962fcf0cde8fcd196e1df666707/client/controller.js#L41-L49
- https://github.com/Financial-Times/n-tracking/blob/bfb23d49ea5e872a3a4ff5b3c9f9253336f9599a/src/client/trackers/realUserMonitoringForPerformance.js#L42-L47

andygout:
Re. [`next-subscribe/client/controller.js`](https://github.com/Financial-Times/next-subscribe/blob/master/client/controller.js#L41-L49), the code is enquiring and interacting with `document` via [`this.document`](https://github.com/Financial-Times/next-subscribe/blob/master/client/controller.js#L12), assigned its value from `this.window.document`, with `this.window` having been assigned from a constructor argument of `window`.

However, `ready-state` is not looking for `this.document`, but rather `document`, and so throws `ReferenceError: document is not defined` when it makes any such enquiry: https://github.com/soenkekluth/ready-state/blob/master/ready-state.js#L10.

Options would seem to be:
- Make a PR to `ready-state` to enable it to accommodate `this.document` as well as `document` (👎).
- Rewrite `next-subscribe/client/controller.js` so uses direct references to `window` and `document` rather than `this.window` and `this.document`, but as a class (`Controller`) this feels like the way it should be.
- Leave it as it is without introducing the `ready-state` package.

i-like-robots:
I can only assume it's done the way it is to make mocking out global methods simpler.

It's not completely necessary to adjust this so I wouldn't spend to much time on it. If it's not a straight swap then I'd move on.

andygout:
Re. [`n-tracking/master/src/client/trackers/realUserMonitoringForPerformance.js`](https://github.com/Financial-Times/n-tracking/blob/master/src/client/trackers/realUserMonitoringForPerformance.js#L42-L47), PR is here: https://github.com/Financial-Times/n-tracking/pull/40.

PR #744, opened 2020-02-04
feat: allow current path override

alexnaish:
Allows the passing in of a `getCurrentPath` function to allow customising of the path that is decorated onto the navigation links.


alexnaish:
Sorry - wrong branch!

PR #745, opened 2020-02-04
feat: allow passing in of getCurrentPath option

alexnaish:
Allows the passing in of a `getCurrentPath` function to allow customising of the path that is decorated onto the navigation links.


i-like-robots:
https://github.com/storybookjs/storybook/issues/9749 has now been fixed

PR #746, opened 2020-02-05
Add renderJsxComponent Handlebars helper

andygout:
Adds a Handlebars helper - `renderJsxComponent` - to those included in `dotcom-server-handlebars`, which does as described in that package's README:

> Outputs a JSX component specified by its local route (relative to the root of the app consuming `dotcom-server-handlebars`) or the package name of the node module, with specification of a named export if required, as well as any other props.

> Example:

```hbs
{{{renderJsxComponent localRoute="views/components/ComponentWithHbsOutput" title="This is a React component"}}}
{{{renderJsxComponent package="@financial-times/dotcom-ui-header" namedExport="LogoOnly" variant="large-logo"}}}
```

I have also updated `examples/combined-hbs-jsx` to:
- use this new Handlebars helper (and remove its own one).
- add a new use case to reflect its ability to render components sourced both locally and from third party dependencies.
- revised the wording to improve clarity (see below screenshots).

#### TODO:
- [x] Get feedback on own PR comments and address.
- [x] Potentially import requisite styles to make visually clear that 'Financial Times' is actually a full JSX component (imported from `dotcom-ui-header`) rather than just a bit of random unstyled text. **Update:** Achieved in this PR: https://github.com/Financial-Times/dotcom-page-kit/pull/750.
- [x] Test use case: `next-myft-page` (HBS) consuming `n-profile-ui` (JSX).
- [ ] Test for new helper in `dotcom-server-handlebars/src/__test__/__fixtures__/helpers.spec.ts`.

#### HBS to JSX
hbs-to-jsx

#### JSX to HBS
jsx-to-hbs


packages/dotcom-server-handlebars/src/helpers/renderJsxComponent.ts
```diff
@@ -0,0 +1,36 @@
+import path from 'path'
+import React from 'react'
+import ReactDOMServer from 'react-dom/server'
+
+export default function renderJsxComponent({ hash }) {
+  let modulePath
+
+  if (hash.hasOwnProperty('package')) {
+    modulePath = path.join(process.cwd(), `node_modules/${hash.package}`)
```
andygout: Can `require.resolve()` be used?

Attempts to use it produced: `Error: Cannot find module 'node_modules/@financial-times/dotcom-ui-header'`.

packages/dotcom-server-handlebars/src/helpers/renderJsxComponent.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  if (!modulePath) {
+    throw new Error('You must specify a "package" or "localRoute" argument to load a module')
+  }
+
+  const importedModule = require(modulePath)
+
+  const Component = hash.hasOwnProperty('namedExport')
+    ? importedModule[hash.namedExport]
+    : importedModule.default
```
andygout: Why is `default` required? It is not required in x-dash: https://github.com/Financial-Times/x-dash/blob/master/packages/x-handlebars/index.js#L38.

packages/dotcom-server-handlebars/src/helpers/renderJsxComponent.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  const props = { ...this, ...hash }
+
+  if (props.hasOwnProperty('_locals')) {
```
andygout: `this` will include `_locals`, meaning this error will always fire. Should we delete that property instead of throwing the error based on its presence?

packages/dotcom-server-handlebars/src/helpers/renderJsxComponent.ts
i-like-robots: `require.resolve(hash.package)` should be fine

packages/dotcom-server-handlebars/src/helpers/renderJsxComponent.ts
```diff
@@ -0,0 +1,36 @@
...
+export default function renderJsxComponent({ hash }) {
+  let modulePath
+
+  if (hash.hasOwnProperty('package')) {
+    modulePath = path.join(process.cwd(), `node_modules/${hash.package}`)
+  }
+
+  // localRoute is relative to root of app consuming dotcom-server-handlebars.
+  if (hash.hasOwnProperty('localRoute')) {
+    modulePath = path.resolve(process.cwd(), `./${hash.localRoute}`)
```
i-like-robots: `path.resolve(hash.localRoute)` is the same =]

packages/dotcom-server-handlebars/src/helpers/renderJsxComponent.ts
i-like-robots: I definitely think we should check!

This can be done quite simply:

```js
  const Component = hash.hasOwnProperty('namedExport')
    ? importedModule[hash.namedExport]
    : importedModule && importedModule.__esModule ? importedModule['default'] : importedModule
```

i-like-robots: Reference: https://github.com/Financial-Times/dotcom-page-kit/blob/a25325d2e534895ac386c508af43a50e9ccf4b56/packages/dotcom-server-react-jsx/src/interopRequire.ts

(This is exactly what TypeScript, Babel, et al all do)

packages/dotcom-server-handlebars/src/helpers/renderJsxComponent.ts
i-like-robots: `this` should only include `_locals` at the root/0 depth level.

There is more info about it here: https://github.com/Financial-Times/x-dash/pull/396

packages/dotcom-server-handlebars/src/helpers/renderJsxComponent.ts
```diff
@@ -0,0 +1,36 @@
+import path from 'path'
+import React from 'react'
+import ReactDOMServer from 'react-dom/server'
+
+export default function renderJsxComponent({ hash }) {
```
i-like-robots: I think `renderReactComponent` would be more accurate as this is not generic.

packages/dotcom-server-handlebars/src/helpers/renderJsxComponent.ts
i-like-robots: Although... because linked dependencies are not flattened you may need to make the header package a direct dependency of the demo app if you're currently consuming via the layout component.

packages/dotcom-server-handlebars/src/helpers/renderJsxComponent.ts
andygout: Yup, `this` does only include `_locals` at the root level but that is the thing that will always cause the error to throw. Think I have missed something.

Simplified, `this` is `{ _locals: {} }`

And `hash` is `{ foo: 'bar' }`

Meaning `const props = { ...this, ...hash }` leaves `props` as `{ _locals: {}, foo: 'bar' }`, meaning `props.hasOwnProperty('_locals')` will always evaluate to true, and always throw the error.

Are you saying we should include this check `this === data.root ? {} : this` that was removed in the referenced PR?

packages/dotcom-server-handlebars/src/helpers/renderJsxComponent.ts
andygout: I _think_ the error might be thrown because the node module it is trying to require is a symlink rather than a directory.

Screenshot 2020-02-05 at 13 24 18

[This discussion](https://github.com/nodejs/node/issues/18009) seems to propose adding an argument to `require.resolve()` to recognise and handle dependencies that are symlinks, but the [Node.js docs for that module](https://nodejs.org/api/modules.html#modules_require_resolve_request_options) indicate that no such feature was ever added.

andygout: Nope, I have disproved this suspicion by re-installing node modules having changed `dependencies` from:
`"@financial-times/dotcom-ui-header": "file:../../packages/dotcom-ui-header"`
to:
`"@financial-times/dotcom-ui-header": "^0.5.8"`
which produces same error.

andygout: This works:
- ``require(path.resolve(`node_modules/${hash.package}`))``

These do not:
- ``require.resolve(`node_modules/${hash.package}`)``
- `require.resolve(hash.package)`

Will keep reading until I full understand what is going on and why `require.resolve()` is not working as expected.

packages/dotcom-server-handlebars/src/helpers/renderJsxComponent.ts
andygout: Fixed in [this commit](https://github.com/Financial-Times/dotcom-page-kit/pull/746/commits/87955298635169ad7f9ac628f062dfd91dd3d93e).

packages/dotcom-server-handlebars/src/helpers/renderJsxComponent.ts
andygout: Fixed in [this commit](https://github.com/Financial-Times/dotcom-page-kit/pull/746/commits/c04c77832aeea9107ad6451a1443e5c1a9681d20), although Prettier had made it pretty ugly…

packages/dotcom-server-handlebars/src/helpers/renderJsxComponent.ts
andygout: Fixed in [this commit](https://github.com/Financial-Times/dotcom-page-kit/pull/746/commits/e603923ae5ab1acde07b5a89b8b6c344adfc59e1).

packages/dotcom-server-handlebars/src/helpers/renderJsxComponent.ts
andygout: Fixed in [this commit](https://github.com/Financial-Times/dotcom-page-kit/pull/746/commits/a79f15f4038f6b153726209e400f500647e0e929).

packages/dotcom-server-handlebars/src/helpers/renderReactComponent.ts
```diff
@@ -0,0 +1,38 @@
...
+export default function renderReactComponent({ hash }) {
+  let modulePath
+
+  if (hash.hasOwnProperty('package')) {
+    modulePath = path.resolve(`node_modules/${hash.package}`)
+  }
+
+  // localRoute is relative to root of app consuming dotcom-server-handlebars.
+  if (hash.hasOwnProperty('localRoute')) {
+    modulePath = path.resolve(hash.localRoute)
```
andygout: Rename `localRoute` -> `localPath`.

packages/dotcom-server-handlebars/src/helpers/renderJsxComponent.ts
andygout: Agreed offline that we should remove this check - done in [this commit](https://github.com/Financial-Times/dotcom-page-kit/pull/746/commits/8083fe47cdd7eb1490adb0c95c579185179a533e).

packages/dotcom-server-handlebars/src/helpers/renderReactComponent.ts
andygout: Fixed in [this commit](https://github.com/Financial-Times/dotcom-page-kit/pull/746/commits/c55273cfa0c239efd490dcabeeef6815dc299d21).

andygout:
#### Test use case:
`next-myft-page` (`page-kit-migration` branch)

With `npm link`s to:
- this branch
- `n-profile-ui` (`chore/handlebars-to-jsx` branch).

`views/universal/alerts/digest.html` (top of that file):

```diff
{{>n-profile-ui/templates/subheading
	text='myFT Email Digest'
	hideLink=true }}

+{{{renderReactComponent
+	package="@financial-times/n-profile-ui"
+	namedExport="Subheading"
+	text='myFT Email Digest'
+	hideLink=true
+}}}
```

#### Subheading successfully rendered using both Handlebars partial template and React component:
Screenshot 2020-02-07 at 09 49 02

Screenshot 2020-02-07 at 10 11 44

andygout:
Changes on other branches mean examples will actually look like this:

#### `hbs-to-jsx`
final-hbs-to-jsx

#### `jsx-to-hbs`
final-3456_jsx-to-hbs

PR #747, opened 2020-02-06
Uppercase README filenames

andygout:
In some directories (e.g. `packages/dotcom-server-handlebars/src/helpers`), having the README filename in lower case (`readme.md`) makes it very difficult to differentiate them from the bulk of non-document files.

This PR uppercases all README filenames so they are uniform.


i-like-robots:
Please note that there are quite a few links to between readme files that will need updating as well as well 🔍

PR #748, opened 2020-02-06
Apply Prettier

andygout:
I think these Prettification changes must have slipped through without the Git hook (that runs `npm run prettier`) being applied (i.e. commits made via GitHub GUI or commit made with `--no-verify`).

Each time I make a commit, `npm run prettier` gets run as the Git hook and I have to then remember to reset the changes it has applied to prevent them from being included in my next commit (where they would not be relevant to that specific commit).

So it seems to make sense to give them their own little commit here where they get applied and stop causing this problem.

### DIFF
--- a/examples/ft-ui/__test__/logged-in-user.test.js
+++ b/examples/ft-ui/__test__/logged-in-user.test.js
@@ -7,11 +7,12 @@ describe('examples/ft-ui', () => {
   describe('Header link elements', () => {
     it('renders the expected loggin-in user Header link elements', async () => {
       await expect(page).toMatchElement('.o-header__top-column--right a[href="/myft"]', { text: 'myFT' })
-      await expect(
-        page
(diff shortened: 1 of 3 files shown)


apaleslimghost:
fwiw, [`lint-staged` runs `git add`](https://github.com/Financial-Times/dotcom-page-kit/blob/master/package.json#L33), which should include the prettier changes in the current commit. we should all also set up a Prettier editor plugin with format-on-save, imo that's the most reliable way of running it

apaleslimghost:
ah, i see what the problem is. the git hook runs Prettier on _all files_, but then `git add` only on the _previously changed files_. we should only format the files that have changed, which is the intended purpose of `lint-staged`

i-like-robots:
It's mildly interesting that the Prettier rules can change in small ways over time, so code that was conformant when it was written can be changed many months later.

PR #749, opened 2020-02-06
change lint-staged rules to only format staged files

apaleslimghost:
previously, if files had slipped through without being formatted, they could be changed in later, unrelated commits, but not written to that commit by `git add`.

this makes sure that only the files staged in the commit get formatted, preventing changes to unrelated files going uncommitted.

this works because `lint-staged` passes the list of staged files as arguments to the commands.

see #748


apaleslimghost:
i'll wait to merge until @i-like-robots has seen this i think, maybe there was a reason for doing it the other way?

PR #761, opened 2020-02-17
Custom contexts

oliverturner:
## What

This is a speculative PR to allow consuming apps to take advantage of PageKit's `AppContext` infrastructure.

## Why

Some apps (e.g. Globetrotter) are overloading `appContext` with their own properties in order to pass config options from server to client.

## How

The proposal is to expose AppContext's JSON embedding and retrieving features to additional custom slots.

**Criteria:**
- Minimal and non-invasive
- Backwards-compatibility

### DIFF
--- a/packages/dotcom-ui-app-context/src/__test__/AppContext.spec.ts
+++ b/packages/dotcom-ui-app-context/src/__test__/AppContext.spec.ts
@@ -8,7 +8,10 @@ const fakeContext = {
     product: 'next',
     abTestState: 'someCohort:on',
     isProduction: true
-  }
+  },
+  customContext1: undefined,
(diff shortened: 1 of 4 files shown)


pixelandpage:
Thanks for opening this! 

Given that the problem you've outlined is that folks are adding properties to appContext when we don't want them to I'm not sure that extending appContext is the ideal solution. If we can think of another approach that lets us preserve appContext as a simple data-schema that might solve the problem more neatly.

I think @i-like-robots had an idea around building a generic dotcom-ui-client-embed package (name TBC 😂) which could become our primary mechanism for making data accessible to the client 🤷‍♀ Do you think that would be feasible?

PR #762, opened 2020-02-17
Add comment which lists all items in the default set

JakeChampion:


pixelandpage:
Out of interest, how will we know if this list changes? Does that ever happen?

JakeChampion:
@pixelandpage The default list is a source of confusion for many users of polyfill.io, I've chosen to never update it and instead recommend that people explicitly list what they require.

JakeChampion:
I'll leave this PR for one of the dotcom-page-kit team to merge and release if that is okay

i-like-robots:
Thanks @JakeChampion !

PR #764, opened 2020-02-18
Use Dart Sass for Sass implementation

andygout:
Fixes https://github.com/Financial-Times/dotcom-page-kit/issues/678.

Switches from using LibSass to Dart Sass for Sass implementation.

Work was validated by running `npm run build` in `./examples/ft-ui` having added the below lines to its `/client/main.scss` file.

```scss
@if true {
  $not-already-defined: true !global;
}
```

Dart Sass throws a warning if a global variable is defined outside of global scope, which LibSass doesn't.

To ensure warnings are enabled you can add to the `main.scss` file the below line which outputs a warning.

```scss
@warn 'warning text';
```

#### Before (i.e. LibSass)
before

---

#### Before (i.e. Dart Sass)
after

#### TODO:
- Testing of this Page Kit branch with consuming apps to ensure that `n-` components are spec compliant.

- [x] `next-static`: https://github.com/Financial-Times/next-static/pull/134. 
- [x] `next-search-page`: no changes required.
- [x] `next-video-page`: no changes required.
- [x] `next-errors`: no changes required.
- [x] `next-epaper`: no changes required.
- [x] `next-tour-page`: https://github.com/Financial-Times/next-tour-page/pull/261.
- [x] `next-graphics-page`: https://github.com/Financial-Times/next-graphics-page/pull/249.
- [x] `next-profile`: https://github.com/Financial-Times/next-profile/pull/292.
- [x] `next-syn-list`: no changes required.
- [x] `next-product`: https://github.com/Financial-Times/next-product/pull/867.
- [x] `next-b2b-prospect`: no changes required.
- [x] `next-retention`: no changes required.
- [x] `next-subscribe`: no changes required.
- [x] `next-stream-page`: no changes required.
- [x] `next-front-page`: no changes required.
- [x] `next-myft-page`: no changes required.
- [x] `next-article`: https://github.com/Financial-Times/next-article/pull/3691.
- [x] `next-control-centre`: no changes required.

#### References:
- [Sass: Dart Sass](https://sass-lang.com/dart-sass).
- [GitHub: webpack-contrib/sass-loader - Options](https://github.com/webpack-contrib/sass-loader#user-content-options).

PR #767, opened 2020-02-20
Replace use of node-sass in Storybook with Dart Sass

i-like-robots:
This change follows #764 to ensure consistency with the Page Kit Webpack configuration.


.storybook/webpack.config.js
```diff
@@ -50,6 +50,7 @@ module.exports = ({ config }) => {
       {
         loader: require.resolve('sass-loader'),
         options: {
+          implementation: require('sass'),
```
andygout: Maybe add comment: ``// Prefer `dart-sass`.`` as there is for this same line in `packages/dotcom-build-sass/src/plugin.ts` to make its purpose clear given its cryptic name.

PR #768, opened 2020-02-21
Update dotcom-ui-layout README description

andygout:


i-like-robots:
I think this description actually needs updating because it does rather more than it makes out.

andygout:
How is:
> This package composes the basic page layout, including selection of the header and footer variants, and insertion of elements that prepend/append them.

PR #773, opened 2020-02-26
add missing react dependency to dotcom-server-handlebars

apaleslimghost:
this is working as-is and not breaking tests because most things that depend on dotcom-server-handlebars also depend on one of the `dotcom-ui-` packages, which have `react` as a dependency, which gets flattened


apaleslimghost:
hold the fort, it also needs `react-dom`

PR #775, opened 2020-03-06
Improve running examples

apaleslimghost:
run jest directly from top level (overriding the `roots` option on the command line) instead of using athloi to iterate over every example and run jest in the subfolder. this makes the cli output nicer and is slightly faster (can use jest's parallel/async test running)


apaleslimghost:
don't think i'll be able to get this to work, because of the separate `jest-puppeteer.config.js` files required for each example. can't find a way around it

PR #777, opened 2020-03-13
Refactor page-kit-cli into a Webpack configuration builder

### LINKED ISSUE #528: Page Kit as a Webpack configuration builder
It has become apparent that the current Page Kit CLI is too complex for our needs and is leading to some confusion about the scope of the project. Because Page Kit provides its own CLI tool it is also worrying users that we are replacing one esoteric system with another.

At its core the current Page Kit CLI enables a complex Webpack configuration to be composed from multiple separate packages. It also makes accessing and modifying specific parts of the configuration simpler and more robust.

However, as discussed in #133 I would like to make it possible to use Webpack's own CLI like this:

```js
// webpack.config.js
module.exports = pageKit(options);
```

Taking this one step further... this would mitigate the need to maintain, document, and support an additional tool. I would also hope that by focusing on the configuration aspect it removes any worries about straying too far from "standard" tools.

This issue supersedes #58 and #123 and #133.

apaleslimghost:
# This PR is part 1 of 2, and should not be reviewed in isolation. Part 2: #778

together with #778, this closes #528.

**summary of changes**:

- removes all CLI-related parts of `dotcom-page-kit-cli`
- exposes `getWebpackConfig` as its main export
- renames it to `dotcom-build-webpack-config`

**why we're doing this**: in FT.com we have a history of building custom tooling and letting it rot for several years leaving nobody able to support it. `page-kit-cli` has a complex architecture, and while it was originally intended to support tasks other than `build`, they never materialised, leaving a poorly-understood wrapper around `webpack-cli` whose only benefit is to allow people to customise the configuration in ways we can't support.

tl;dr less custom stuff, more developer confidence

## how to test this PR

sort answer: don't. long answer: test #778 instead. if it works there it'll work here; if it doesn't work there we probably can't fix it here. you can run webpack in the examples and check the output there if you like but testing #778 will be simpler.

## how to review this PR

- the main entry point for the new non-CLI package is [`packages/dotcom-build-webpack-config/src/operations/getWebpackConfig.ts`](https://github.com/Financial-Times/dotcom-page-kit/pull/777/files#diff-ddfed7352f76cca6c205632a99cba2b3), but this hasn't actually changed very much [from `master`](https://github.com/Financial-Times/dotcom-page-kit/blob/b888454/packages/dotcom-page-kit-cli/src/operations/getWebpackConfig.ts). the main things that _have_ changed there are formatting and names of things (e.g. since it's no longer a CLI it doesn't make sense to have a class called `CliContext`)
- have a look at what's been removed between `dotcom-page-kit-cli` and `dotcom-build-webpack-config`, it should be all CLI-related stuff, and everything should still fit together. note that Github hasn't picked up on some moved (not deleted) files so check filenames
- the changes in the other `dotcom-build-` plugins are to do with removing access to settings from `dotcom-page-kit-cli` which we're mostly making unconfigurable (it doesn't make sense to let people change the `babel-preset-env` settings, that'll just cause different bundle contents to other apps for no reason)
- the examples have been migrated from `page-kit-cli` to the `getWebpackConfig`-based Webpack config format. they should all work.
- because this is an intermediate step, some of the documentation is incomplete, as it's pointless documenting something that's immediately going to be refactored in the next PR without ever reaching users. please review the finished documentation in #778.


examples/building-sass-files/public/manifest.json
```diff
@@ -0,0 +1,10 @@
+{
+  "entrypoints": {
+    "styles": {
+      "css": [
+        "styles.css"
+      ]
+    }
+  },
+  "styles.css": "styles.css"
+}
```
i-like-robots: I don't think the contents of the `public` directory should be here

examples/dynamic-imports/webpack.config.js
```diff
@@ -0,0 +1,15 @@
...
+const { getWebpackConfig } = require('@financial-times/dotcom-build-webpack-config')
+const js = require('@financial-times/dotcom-build-js')
+
+module.exports = getWebpackConfig(
+  {
+    entry: {
+      scripts: './client/main.js'
+    },
+    output: {
+      publicPath: '/public/'
```
i-like-robots: Did you mean `path.resolve('public')`?

apaleslimghost: this is `publicPath`, it's for a URL not a file path

examples/building-sass-files/public/manifest.json
apaleslimghost: 🤔

apaleslimghost: fixed this, example was expecting a `dist` folder, webpack is outputting to `public`

apaleslimghost:
we can ignore the snyk failures, they happen every time a package is deleted

packages/dotcom-build-js/src/plugin.ts
```diff
@@ -25,7 +25,8 @@ export function plugin(userOptions: PluginOptions = {}) {
     scriptsRule.test.push(/\.(ts|tsx)$/)
   }
 
-  function addBabelPreset({ cli, resource: babelConfig }: HandlerArgs) {
-    babelConfig.presets.push(babelPreset(options, cli))
+  function addBabelPreset({ context, resource: babelConfig }: HandlerArgs) {
+     console.log(arguments)
```
pixelandpage, line 29: Do we want to remove this console.log?

apaleslimghost: yes we do

apaleslimghost: good catch

i-like-robots:
This looks good, I'll quickly check it out and run the examples as well

PR #778, opened 2020-03-13
Remove pluggable from dotcom-build- packages

### LINKED ISSUE #528: Page Kit as a Webpack configuration builder

apaleslimghost:
# This PR is part 2 of 2, and should not be reviewed in isolation. Part 1: #777

together with #777, this closes #528.

**summary of changes**:

- removes `dotcom-pagekit-pluggable`
- refactors all Pluggable plugins into Webpack plugins
- refactors `dotcom-build-webpack-config` into a Webpack plugins

**still to finish**:

- [x] migrate the examples
- [x] switching `thePlugin.plugin()` syntax to `new ThePlugin()`?
- [x] renaming the package from `dotcom-build-webpack-config`?
- [x] migration guide

**why we're doing this**:

Pluggable, although well-documented, is a complex and poorly-understood package. it's intended to let people make deep changes to the Page Kit Webpack configuration, but we don't want to support that, as our configuration is intended as a common baseline for client-side assets for all user-facing apps. we can't guarantee performance and cachability if consumers can change the config.

Webpack plugins are perfectly capable of modifying the compiler's configuration as it loads, so this PR refactors the Pluggable plugins into Webpack plugins. this leaves the entry point, `dotcom-build-webpack-config`, with nothing to do other than set some config, so it gets turned into a Webpack plugin as well.

because there's now no single Webpack entry point, I've added a guide to using Webpack in `docs/guides`. i'll also write a guide to migrate from `page-kit-cli` to `webpack-cli`.

tl;dr less custom stuff, more developer confidence


apaleslimghost:
still to consider:

- [ ] migration guide
- [ ] renaming the package from `dotcom-build-webpack-config`?

apaleslimghost:
cool i fixed the snyk problem

.eslintrc.js
```diff
@@ -3,26 +3,26 @@ module.exports = {
   env: {
     node: true,
     browser: true,
-    es6: true
+    es6: true,
```
apaleslimghost: i don't know why prettier has suddenly added trailing commas everywhere

apaleslimghost:
this is ready to review, the migration guide will be in the wiki, not part of this PR

apaleslimghost:
draft migration guide available here: https://github.com/Financial-Times/dotcom-page-kit/wiki/Upgrading-from-0.6.x-to-0.7.x

.eslintrc.js
i-like-robots: The default setting changed in Prettier 2.0, perhaps the version provided by your editor changed? I have amended the rules in #783 to avoid this.

apaleslimghost: oh, i had to update it in the PR to support the Typescript 3.8 `import type` syntax i started using. let's merge #783 and i'll rebase

i-like-robots: Done 👍

examples/kitchen-sink/webpack.config.js
```diff
@@ -1,32 +1,31 @@
...
         // All Bower components are installed at the repo root
-        path.resolve('../../bower_components')
-      ]
+        path.resolve('../../bower_components'),
+      ],
     }),
-    js.plugin()
-  ]
-)
+    new PageKitJsPlugin(),
```
pixelandpage: Super minor, but could we move the JS plugin above the sass one? It might be easier to see that it's part of the same array, with sass being the only one to have extra config options.

packages/dotcom-build-js/README.md
```diff
@@ -1,11 +1,9 @@
 # @financial-times/dotcom-build-js
```
pixelandpage, line 1: It would be good to document the use of `babel` in this plugin as it's been moved out of the base styles package now and it's a reasonably important piece of the Page Kit puzzle.

packages/dotcom-build-base/src/index.ts
```diff
@@ -0,0 +1,54 @@
+import type webpack from 'webpack'
+import path from 'path'
+import { CleanWebpackPlugin } from 'clean-webpack-plugin'
+import CompressionPlugin from 'compression-webpack-plugin'
+import ManifestPlugin from 'webpack-assets-manifest'
+
+export class PageKitBasePlugin {
+  apply(compiler: webpack.Compiler) {
```
pixelandpage, line 8: I don't really understand the `apply(compiler)` bit but I'm guessing it's a standard Webpack thing. Do you think you could you explain what it's doing for my benefit?

apaleslimghost: it's a standard Webpack thing. when Webpack loads, it goes through the `plugins` array and calls their `apply` method with the instance of the compiler (the compiler is what you get if you call `webpack()` programatically instead of using the CLI). 

- https://webpack.js.org/contribute/writing-a-plugin/#basic-plugin-architecture
- https://github.com/webpack/webpack/blob/webpack-4/lib/webpack.js#L46-L54

apaleslimghost: so, this isn't "doing" anything, it's defining what Webpack should call when loading

packages/dotcom-build-js/README.md
apaleslimghost: just to be clear, do you mean reinstate some documentation that's missing from this branch, or write new documentation for babel? the only babel docs i can see that have been remove are around the hooks for customising the config, which aren't applicable any more

docs/guides/building-client-side-assets-with-webpack.md
```diff
@@ -0,0 +1,138 @@
...
+Install the Webpack CLI, and the base Page Kit Webpack config plugin:
+
+```
+npm install --save-dev webpack webpack-cli @financial-times/dotcom-build-base
+```
+
+Create a `webpack.config.js` file in the root of your app. This is the default [Webpack configuration file]. You should add an `entry` option pointing at your client-side entry point (usually `client/main.js`), and `plugins` array initialising the Page Kit config plugin:
+
+```js
+const { PageKit$1Plugin } = require('@financial-times/dotcom-build-base')
```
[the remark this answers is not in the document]
apaleslimghost: this is a find/replace error. it's happened before to me, might be a bug in atom 🤔

packages/dotcom-build-js/README.md
pixelandpage: Let's not worry about adding new documentation but an indication in this readme that this is where the babel happens would probably be useful. Babel docs are pretty accessible if folks need extra context.

packages/dotcom-build-code-splitting/src/index.ts
```diff
@@ -1,4 +1,100 @@
...
+          name: 'webpack-runtime'
+        },
+        splitChunks: {
+          chunks: 'all'
+        },
+        // We're going to implement our own algorithm so don't double effort
+        moduleIds: false,
+        chunkIds: 'named'
+      }
+    }
```
i-like-robots, line 28: I think this could be simplified in future, the `cacheGroups` could be specified right here to avoid the need to deep merge everything later

packages/dotcom-build-js/src/babel.ts
```diff
@@ -1,46 +1,65 @@
...
     // them down and actually result in fewer function calls overall!
     // <https://github.com/Financial-Times/dotcom-page-kit/issues/576>
     helpers: false
-  })
+  }
 
   const config = {
+    // By default Babel assumes all source code is ESM so force it to check for CJS
+    sourceType: 'unambiguous',
+    babelrc: true,
```
i-like-robots, line 33: Should this be false? The readme says "Other methods of configuring Babel are not supported" does that impy not possible or you're on your own?

docs/guides/building-client-side-assets-with-webpack.md
```diff
@@ -0,0 +1,141 @@
...
+## Integration with FT.com build tools
+
+Apps using [`n-gage`](https://github.com/financial-times/n-gage) should add `build` and `build-production` targets to their `Makefile`:
+
+```makefile
+build:
+	webpack --mode=development
+
+build-production:
+	webpack # defaults to production mode
```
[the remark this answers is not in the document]
apaleslimghost: they're tabs :)

packages/dotcom-build-js/src/babel.ts
apaleslimghost: this hasn't changed from the old config. i'm implying they're on their own

docs/guides/building-client-side-assets-with-webpack.md
```diff
@@ -0,0 +1,141 @@
...
+Install the Webpack CLI, and the base Page Kit Webpack config plugin:
+
+```
+npm install --save-dev \
+   webpack \
+   webpack-cli \
+   @financial-times/dotcom-build-base
+```
+
+Create a `webpack.config.js` file in the root of your app. This is the default [Webpack configuration file]. You should add an `entry` option pointing at your client-side entry point (usually `client/main.js`), and `plugins` array initialising the Page Kit config plugin:
```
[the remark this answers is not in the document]
apaleslimghost: i've just removed the `you should`

examples/ft-ui/__test__/anonymous-user.test.js
```diff
@@ -16,11 +16,12 @@ describe('examples/ft-ui', () => {
...
-        page
-      ).not.toMatchElement('.o-header__nav-list--right a[href="https://www.ft.com/myaccount"]', {
-        text: 'Settings & Account'
-      })
+      await expect(page).not.toMatchElement(
+        '.o-header__nav-list--right a[href="https://www.ft.com/myaccount"]',
+        {
+          text: 'Settings & Account'
+        }
+      )
```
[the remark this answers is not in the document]
apaleslimghost: this is "correct" prettier formatting. it doesn't care about consistency between different lines, just the current line length :/

PR #781, opened 2020-03-18
Update Changelog for v0.6.3

andygout:


changelog.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## 0.6.3
+
+- CLI package: Modify WebpackError to handle undefined `stats` value.
+- Header package: Relabel "Account Settings" as "Settings and Accounts" to be more consistent with mobile app terminology; redirect old account settings page to new Manage My Account page.
+- Handlebars package: Add missing `react` and `react-dom` dependencies.
+- Multiple packages: Remove incremental compiler option from root TS config to prevent conflict
```
i-like-robots: This is an internal detail that is not relevant to consumers so there's no need to mention it

changelog.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## 0.6.3
+
+- CLI package: Modify WebpackError to handle undefined `stats` value.
+- Header package: Relabel "Account Settings" as "Settings and Accounts" to be more consistent with mobile app terminology; redirect old account settings page to new Manage My Account page.
```
i-like-robots: This is an internal detail, the only thing that changed was fixtures and tests so not necessary for consumers to know about

PR #786, opened 2020-03-31
Updating version of o-fonts-assets

chivchila:
Origami team has recently updated fonts with a new font https://github.com/Financial-Times/o-fonts-assets/releases/tag/v1.4.0. Note that this PR isn't shipping any new fonts, only updating the o-asset-fonts links

### DIFF
--- a/packages/dotcom-ui-base-styles/src/lib/fontFaces.ts
+++ b/packages/dotcom-ui-base-styles/src/lib/fontFaces.ts
@@ -1,3 +1,5 @@
+// NOTE: please also update ./styles.css when updating the URLs defined below. They
+// need to stay in-sync to ensure our CSS output matches the resource hints we set.
 export const fontFaceURLs = [
   'https://www.ft.com/__origami/service/build/v2/files/o-fonts-assets@1.4.1/MetricWeb-Regular.woff',
   'https://www.ft.com/__origami/service/build/v2/files/o-fonts-assets@1.4.1/MetricWeb-Semibold.woff',
--- a/packages/dotcom-ui-base-styles/src/lib/fontFaces.ts
+++ b/packages/dotcom-ui-base-styles/src/lib/fontFaces.ts
@@ -1,6 +1,6 @@
 export const fontFaceURLs = [
-  'https://www.ft.com/__origami/service/build/v2/files/o-fonts-assets@1.3.2/MetricWeb-Regular.woff',
-  'https://www.ft.com/__origami/service/build/v2/files/o-fonts-assets@1.3.2/MetricWeb-Semibold.woff',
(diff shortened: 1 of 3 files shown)


i-like-robots:
It looks like `o-fonts` references a different URL so these resource hints don't currently match those referenced by our CSS... https://github.com/Financial-Times/o-fonts/blob/master/src/scss/_variables.scss#L9

Perhaps we should set this variable as part of the base styles package to ensure they stay aligned @pixelandpage ?

pixelandpage:
> Perhaps we should set this variable as part of the base styles package to ensure they stay aligned

Sounds sensible, we want to make sure we're always pulling through fonts with the right version number. Where would the source of truth be for this?

pixelandpage:
If Origami are willing to export the version number in o-fonts we could pull it in from there 🤔

i-like-robots:
@pixelandpage if we define the `$o-fonts-path` in the base styles package then I _think_ all references to font URLs will live in one place \0/

pixelandpage:
Hmm, I not sure I'm following you completely. We could declare the variable manually in this file, like so:
```
const o-fonts-path = 'https://www.ft.com/__origami/service/build/v2/files/o-fonts-assets@1.4.1';

export const fontFaceURLs = [
	`${o-fonts-path}/MetricWeb-Regular.woff`,
	...
]
```
And then update `o-fonts-path` manually any time we want to bump o-fonts.

i-like-robots:
That's JS and we're in Sass-land =]

pixelandpage:
Ok, let's leave the `o-fonts-path` aside for now and pull that in separately. 

@chivchila, can you update the fonts urls so that they match the version in o-fonts?

i-like-robots:
@notlee are the changes in 400f4f3ff391560bcc860aeb97e196b05b99b9d5 sensible in your opinion? The intention is to centralise the font URLs in one place (across Sass and JS) to ensure they stay in-sync.

notlee:
Yeah, it's not clean but I think it's sensible :)

packages/dotcom-ui-base-styles/src/lib/fontFaces.ts
```diff
@@ -1,6 +1,8 @@
+// NOTE: please also update ./styles.css when updating the URLs defined below. They
```
apaleslimghost, line 1: i'm sure i've written a package at some point to extract variables from sass. can't find it right now. so, it's possible, but i think this is fine for now

apaleslimghost: oh, i didn't write it, but i have used it before https://github.com/XOP/sass-vars-to-js

PR #789, opened 2020-04-03
update CODEOWNERS to @financial-times/platforms team

apaleslimghost:


apaleslimghost:
using the admin button to yeet it into master because it doesn't change the code

PR #791, opened 2020-04-08
NOPS-71: Hide subscribe sticky nav button when screen is small

nickramsbottom:
## Summary
The nav items in the masthead are inline and still showing on small screens causing elements to run into each other. The subscribe button is now hidden on smaller screens.

I've tested this locally and got the desired result but would appreciate someone else testing as well to make sure

After:
image

Before:
incorrect_buttons_aligned_with_masthead

### DIFF
--- a/packages/dotcom-ui-header/bower.json
+++ b/packages/dotcom-ui-header/bower.json
@@ -1,7 +1,7 @@
 {
   "name": "dotcom-ui-header",
   "dependencies": {
-    "o-header": "^8.0.0",
+    "o-header": "^8.2.0",
     "n-topic-search": "^2.0.0",
     "n-ui-foundations": "^4.0.0-beta.2"
   },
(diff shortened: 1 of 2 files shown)


pixelandpage:
This looks good 👍 The tests are failing on Circle but passing for me locally. Are you able to recreate the errors?

pixelandpage:
The CI failures are definitely unrelated to these changes. I'll try to debug it in the morning.

pixelandpage:
@nickramsbottom the Circle errors should be solved now. If you rebase over master the tests should pass 🙂

nickramsbottom:
@pixelandpage Amazing, thank you!

binaryberry:
Screenshot 2020-04-16 at 11 18 57

Looks good to me! I do notice the spacing is still weird in full screen, but that's a separate card.

nickramsbottom:
@binaryberry @pixelandpage thank you both :D

binaryberry:
Looking at the homepage in the dev tools, I can't find a device that has the width at which the problem I mentioned above problem is visible! They all have either a bigger or smaller screen. So I won't create a card for this after all!

PR #793, opened 2020-04-08
Remove tests which expect the length of a function

pixelandpage:
Removes two test expectations which call `toHaveLength()` on Function values. 

I'm not sure what the original intention was but this usage is pretty unexpected (usually only strings and arrays have `length` properties) and it's not reflected in the examples given in the jest docs. https://jestjs.io/docs/en/expect#tohavelengthnumber

Both test conditions seem to be covered by remaining expectations so it should be safe to remove these.

### DIFF
--- a/packages/dotcom-middleware-app-context/src/__test__/index.spec.ts
+++ b/packages/dotcom-middleware-app-context/src/__test__/index.spec.ts
@@ -34,7 +34,6 @@ describe('dotcom-middleware-app-context', () => {
 
   it('returns a request handler function', () => {
     expect(instance).toBeInstanceOf(Function)
-    expect(instance).toHaveLength(3)
   })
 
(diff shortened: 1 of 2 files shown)


i-like-robots:
Just to provide some context, this test was checking the provided callback expected 3 arguments: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/length

However, the test is not very useful and if the interface were to accidentally change it would be caught by other specs in the suite =]

pixelandpage:
Thanks for the context @i-like-robots 😄

i-like-robots:
And this is why it broke: https://github.com/facebook/jest/pull/9626/files#diff-99739ad11655a4534ce3ab4222dbddbc

Issue #794, opened 2020-04-14
Upgrade apps to Page Kit v0.7


andygout:
#### Migration guide
https://github.com/Financial-Times/dotcom-page-kit/wiki/Upgrading-from-0.6.x-to-0.7.x

#### Progress spreadsheet (should also be updated as we go along)
https://docs.google.com/spreadsheets/d/1wOp6suvluEVmD3v8GEmP5jXRzoSpUazSN4G5HHdpQY0/edit?ts=5e173179#gid=0

#### TODO:
- `next-static` (Platforms)
  - [x] https://github.com/Financial-Times/next-static/pull/139
- `next-search-page` (Content Discovery)
  - [x] https://github.com/Financial-Times/next-search-page/pull/383
- `next-video-page` (Content Innovation)
  - [ ] https://github.com/Financial-Times/next-video-page/tree/upgrade-page-kit - appContext: Blocked by https://github.com/Financial-Times/dotcom-page-kit/pull/799
- `next-errors` (Platforms)
  - [x] https://github.com/Financial-Times/next-errors/pull/200
- `next-epaper` (Content Innovation)
  - [x] https://github.com/Financial-Times/next-epaper/pull/99
- `next-tour-page` (Acquisition)
  - [x] https://github.com/Financial-Times/next-tour-page/pull/269
- `next-graphics-page` (Content Innovation)
  - [x] https://github.com/Financial-Times/next-graphics-page/pull/255
- `next-profile` (Accounts)
  - [x] https://github.com/Financial-Times/next-profile/pull/380
- `next-syn-list` (Accounts)
  - [x] https://github.com/Financial-Times/next-syn-list/pull/102
- `next-product` (Acquisition)
  - [x] https://github.com/Financial-Times/next-product/pull/882
- `next-retention` (Accounts)
  - [x] https://github.com/Financial-Times/next-retention/pull/423
- `next-subscribe` (Acquisition)
  - [x] https://github.com/Financial-Times/next-subscribe/pull/1055
- `next-stream-page` (Content Discovery)
  - [x] https://github.com/Financial-Times/next-stream-page/pull/2405
- `next-front-page` (Content Discovery)
  - [x] https://github.com/Financial-Times/next-front-page/pull/2321
- `next-myft-page` (Content Discovery)
  - [x] https://github.com/Financial-Times/next-myft-page/pull/2269
- `next-article` (Content Innovation)
  - [x] https://github.com/Financial-Times/next-article/pull/3770
[...]

PR #798, opened 2020-04-16
Capitalise Makefile filename

andygout:
I kept not being able to find the Makefile because I was looking for its name being capitalised.

This creates consistency with how it is written elsewhere throughout Customer Products.


pixelandpage:
Oh I'm not sure about this 😕 What apps are you trying to be consistent with?

andygout:
At least the 17 apps on which we are currently performing Page Kit migrations: https://github.com/Financial-Times/dotcom-page-kit/issues/794.

pixelandpage:
On the one hand, you've made a strong case for consistency! On the other, we don't use this capitalisation pattern for any other root-level files and I don't mind Page Kit breaking with conventions when we have a good reason to. On balance, I can reluctantly accept this based on the fact that it might improve usability, but I'd be interested to know _why_ we capitalise the `Makefile` in our existing repos - this doesn't seem to be backed by any other conventions.

andygout:
Seems the convention is to capitalise (although the reasoning for this comes from the handling of letter casing only on certain operating systems): https://stackoverflow.com/questions/14368941/is-the-convention-for-naming-make-files-to-use-a-capital-m-such-as-makefile/14368985

Having only ever seen it capitalised means my muscle memory kept missing it in lowercase, so this PR was made exclusively from a usability angle.

Happy to merge or close either way.

andygout:
Agreed offline to close.

PR #801, opened 2020-04-22
Revert "Relocate LoadFontsEmbded prior to Stylesheets for minor performance optimisation"

apaleslimghost:
This causes a font loading race condition in Chrome that means sometimes,
although fonts have loaded, we don't remove the classes from the body,
so the fallback fonts are displayed.

This reverts commit 721eb24363a2a26d3809f993e2afd9c01c0bf63b.

https://financialtimes.atlassian.net/browse/NOPS-120


apaleslimghost:
go for it. we should prioritise rolling out the fix to front page and article page but after that i think do it at leisure

PR #770, opened 2020-02-25
Create base style package

i-like-robots:
This new package provides the global document styles for FT.com, including basic typography, normalisation styles, and grid.

The intention of this package is to centralise all of the required pieces for o-typography's progressive font loading and create a placeholder for the future inclusion of styles from `n-ui-foundations`.

It provides 3 parts, firstly the server-side integration:

```jsx
import {
  documentStyles
  LoadFontsEmbed,
  loadCustomFontsClassNames
} from '@financial-times/dotcom-ui-base-styles'

export default (props) => (
  <html className={loadCustomFontsClassNames} style={documentStyles}>
    <head>
      <meta charSet="utf-8" />
      <title>My Amazing Website</title>
      <LoadFontsEmbed />
    </head>
    <body>
      ...
    </body>
  </html>
)
```

Secondly, client-side JS:

```js
import * as baseStyles from '@financial-times/dotcom-ui-base-styles'

baseStyles.init()
```

And lastly, the styles:

```scss
@import '@financial-times/dotcom-ui-base-styles/styles';
```

Resolves #494

PR #803, opened 2020-04-23
fix fonts loading

apaleslimghost:
welp. fun fact: font loading has never worked with Page Kit! "but bren," i hear you ask, "yes they have what the heck?", well:

- by default, we add font loading classes to the `<html>` element, which forces o-typography to use the fallback fonts while fonts load
- there's a separate bit of code to remove the loading classes if the fonts have already loaded; `o-typography` uses a cookie, `o-typography-fonts-loaded`, to signal this
- `dotcom-ui-layout`'s init calls `oTypography.init` on the `.n-layout` element. it's supposed to load the fonts and set the cookie, but it has no effect, because that element doesn't have `data-o-component="o-typography"` [any more](https://github.com/Financial-Times/dotcom-page-kit/pull/770/files#diff-b7d9d78266cd0478e141461d7d6f73a0L83)
- `dotcom-ui-base-style`'s init also calls `oTypography.init` on the default of `<html>`, but that doesn't have the `data-o-component` either, and also no apps actually use this `init`
- almost everybody already has the cookie! because fonts used to load just fine with, and they still have the cookie
- it's only now the cookie has started expiring for people, or in incognito, that people are seeing broken fonts
- previous attempts to fix this seemed to work because developers (me) had the cookie in their browser
- ~recent changes to font loading are a red herring, it's just a timing thing. this is why attempting to go back to a working version of an app didn't seem to do anything~

so, we fix it by calling `oTypography.init` with no args, so it defaults to `<html>`, and adding `data-o-component="o-typography"` to `<html>`

- [x] it's still a bit weird that both `dotcom-ui-layout` and `dotcom-ui-base-styles` call `oTypography.init`. should i update `dotcom-ui-layout` to call `dotcom-ui-base-styles` instead of `o-typography` directly? this would assume that every app using `dotcom-ui-layout` also uses `dotcom-ui-base-styles`

### DIFF
--- a/packages/dotcom-ui-layout/browser.js
+++ b/packages/dotcom-ui-layout/browser.js
@@ -1,6 +1,6 @@
 import * as footer from '@financial-times/dotcom-ui-footer/browser'
 import * as header from '@financial-times/dotcom-ui-header/browser'
-import * as baseStyles from '@financial-times/dotcom-ui-base-styles'
+import * as baseStyles from '@financial-times/dotcom-ui-base-styles/browser'
 // Polyfill for :focus-visible https://github.com/WICG/focus-visible
 // NOTE: v5 of this polyfill is not yet supported by o-normalise
 // https://github.com/WICG/focus-visible/pull/196/files
--- a/packages/dotcom-ui-layout/browser.js
+++ b/packages/dotcom-ui-layout/browser.js
@@ -1,6 +1,6 @@
 import * as footer from '@financial-times/dotcom-ui-footer/browser'
 import * as header from '@financial-times/dotcom-ui-header/browser'
-import oTypography from 'o-typography'
+import * as baseStyes from '@financial-times/dotcom-ui-base-styles'
 // Polyfill for :focus-visible https://github.com/WICG/focus-visible
 // NOTE: v5 of this polyfill is not yet supported by o-normalise
 // https://github.com/WICG/focus-visible/pull/196/files
@@ -8,7 +8,7 @@ import oTypography from 'o-typography'
 import 'focus-visible'
 
 export function init({ headerOptions = {}, footerOptions = {} } = {}) {
-  oTypography.init()
+  baseStyles.init()
   header.init(headerOptions)
   footer.init(footerOptions)
 }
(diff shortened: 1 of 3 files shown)


packages/dotcom-ui-layout/browser.js
```diff
@@ -8,8 +8,7 @@ import oTypography from 'o-typography'
 import 'focus-visible'
 
 export function init({ headerOptions = {}, footerOptions = {} } = {}) {
-  const rootElement = document.querySelector('.n-layout')
-  oTypography.init(rootElement)
+  oTypography.init()
```
i-like-robots, line 11: Looks like this was an oversight on my part in https://github.com/Financial-Times/dotcom-page-kit/pull/770

The intention was for base styles package to provide all font loading related code, including this. For whatever reason this must have gotten lost but I believe the layout package should have imported and used the base styles JS instead: https://github.com/Financial-Times/dotcom-page-kit/blob/master/packages/dotcom-ui-base-styles/browser.js

pixelandpage:
> this would assume that every app using dotcom-ui-layout also uses dotcom-ui-base-styles

I think this is a safe assumption - we should keep the logic in the base styles package and trust that it will be pulled in via layout 🙂

packages/dotcom-ui-layout/browser.js
pixelandpage, line 11: As we're going to be pull this logic in from the base styles package, I'm pretty sure we can remove this line from the layout 🙂

notlee:
>it's supposed to load the fonts and set the cookie, but it has no effect, because that element >doesn't have data-o-component="o-typography"

The `data-o-component` should not be required when explicitly passing an element to the `init` function of an Origami component. This is definitely an inconsistency in o-typography. I'll open a PR to correct that -- we're looking into how to empower other teams to create Origami compatible components, so should stop this kinda thing happening in the future as part of that effort

Like we talked about earlier it would be way better to remove all this [typography js](https://github.com/Financial-Times/o-typography/issues/248#issue-569108606)

apaleslimghost:
@pixelandpage i've updated this so `dotcom-ui-layout` imports `dotcom-ui-base-styles` and calls that

notlee:
>The data-o-component should not be required when explicitly passing an element to the init function of an Origami component.

Turns out I was wrong! Other Origami components do consistently expect the `data-o-component` when passing an element to the `init` function. Not on elements found when passing a selector though...

apaleslimghost:
turns out the [`data-o-component` did used to be on the `n-layout` element](https://github.com/Financial-Times/dotcom-page-kit/pull/770/files#diff-b7d9d78266cd0478e141461d7d6f73a0L83), so it did break in 0.7. that's why it's breaking only now, the cookie has an expiry of one week, and it's just over a week since most apps were migrated.

PR #804, opened 2020-04-27
Add v0.7.3 to changelog.md

andygout:


examples/security-headers/public/manifest.json
```diff
@@ -0,0 +1,22 @@
+{
```
apaleslimghost, line 1: should these files be in this PR?

andygout: 😬 Certainly not. Carried over from another branch. Will remove.

PR #806, opened 2020-05-04
Add customClass to footerOptions

emmalewis:
This Pull Request adds a customClass option to footerOptions so that applications can pass this through layoutProps and have a class render on their footer.

Use case: in next-product we want to remove the top margin from the footer on a single page.

### DIFF
--- a/packages/dotcom-ui-footer/src/__test__/component.spec.tsx
+++ b/packages/dotcom-ui-footer/src/__test__/component.spec.tsx
@@ -27,6 +27,14 @@ describe('dotcom-ui-footer', () => {
     expect(legalFooter.find('.o-footer__brand-logo')).toExist()
   })
 
+  it('renders a custom class on the footer', () => {
+    dataFixture.customClass = 'foo'
+    const footer = mount(<Footer {...dataFixture} />)
+    const legalFooter = mount(<Footer {...dataFixture} />)
+    expect(footer.find('.o-footer').hasClass('foo')).toBe(true)
+    expect(legalFooter.find('.o-footer').hasClass('foo')).toBe(true)
+  })
+
   describe('standard footer', () => {
     it('renders the navigation sections', () => {
       expect(footer).not.toBeEmptyRender()
--- a/packages/dotcom-ui-footer/src/index.tsx
+++ b/packages/dotcom-ui-footer/src/index.tsx
@@ -10,6 +10,7 @@ import {
 
 export type TFooterOptions = {
   theme?: 'dark' | 'light' | string
+  customClass?: string
 }
 
 export type TFooterProps = TFooterOptions & {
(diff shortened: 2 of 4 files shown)


packages/dotcom-ui-footer/src/index.tsx
```diff
@@ -20,7 +21,10 @@ export function Footer(props: TFooterProps) {
   const footerData = props.data.footer.items
   const theme = props.theme ? `${props.theme}` : 'dark'
   return (
-    <footer id="site-footer" className={`o-footer o-footer--theme-${theme}`} data-o-component="o-footer">
+    <footer
+      id="site-footer"
+      className={`o-footer o-footer--theme-${theme} ${props.customClass}`}
```
andygout, line 26: What happens in the event of a `customClass` attribute not being added to `props`?

It seems for `theme` that default values are set on lines 22 and 40 respectively and that maybe the same may be required here.

I may well be wrong but it looks like it might produce: `o-footer o-footer--theme-dark undefined`.

packages/dotcom-ui-footer/src/__test__/component.spec.tsx
```diff
@@ -27,6 +27,14 @@ describe('dotcom-ui-footer', () => {
     expect(legalFooter.find('.o-footer__brand-logo')).toExist()
   })
 
+  it('renders a custom class on the footer', () => {
+    dataFixture.customClass = 'foo'
+    const footer = mount(<Footer {...dataFixture} />)
+    const legalFooter = mount(<Footer {...dataFixture} />)
+    expect(footer.find('.o-footer').hasClass('foo')).toBe(true)
+    expect(legalFooter.find('.o-footer').hasClass('foo')).toBe(true)
+  })
```
andygout, line 36: Might be worth adding a test to see what the entire class value looks like when `customClass` is absent, in relation to other comment about `undefined` possibly being included.

packages/dotcom-ui-footer/src/index.tsx
andygout: I also wonder if it is possible to avoid trailing whitespace when `props.customClass` is absent (maybe by putting classes in an array and `.join`ing them with `' '`).

andygout:
Coming at the problem from another angle, is it possible to change the background colour of the area exposed by the footer top margin to match the content and footer background colour?

It would potentially be a similar fix to this PR.

If this fix is essentially to disguise the pink bar of the background, then we should try to change its background colour rather than its position to its neighbouring components. While there are variants of the header and footer, it still seems that they should have a consistent position in relation to their neighbouring components where possible.

pixelandpage:
Hi Emma, thanks so much for opening this!

I'm a bit concerned about adding an open `customClass` property to the `footerOptions` as it's very open and we don't have a convention for this elsewhere. Setting an option called something like `collapseFooterMargin` on `footerOptions` with a default boolean value would allow you to optionally set a class on this component. We'd need to take care not to add a class of `undefined` while doing this, which I think we would with the current syntax. 

Another option would be to hook directly to `o-footer` in your app to add collapse the `margin-top` value. It's not ideal, but it would keep the logic isolated inside your app without requiring any changes to Page Kit. 

I would also caution against removing this margin though. It's a standard feature across all of our Origami footers and from experience editing Origami properties at this level can cause unintended headaches later when Origami make changes to their components. Would changing the `background-color` be an alternative to this?

emmalewis:
@andygout @pixelandpage Thank you for your comments. Following on from further discussions with Maggie today, and bearing in mind that this is a test for now, I'm going to close this PR and do a work around locally. We may need to revisit this in the future if the test becomes a fixture, in which case it'd be good to set up a meeting and discuss properly. Many thanks!

pixelandpage:
Okie dokie. Thanks for the update @emmalewis 🙂

PR #782, opened 2020-03-19
Fail validation when app context receives an unknown property

i-like-robots:
All app context properties are appended to tracking events. The schema currently allows unknown properties to added for convenience, however this is causing the analytics team problems as they are receiving unexpected data.

This updates the app context schema to disallow unknown properties and reiterates how app context is used in the readme.

We will provide another appropriate mechanism for passing data between the server and client (cc #761).

PR #808, opened 2020-05-05
Add `dotcom-ui-data-embed`

nickcolley:
Continued in this pull request: https://github.com/Financial-Times/dotcom-page-kit/pull/812


packages/dotcom-ui-flags/src/components/FlagsEmbed.ts
```diff
@@ -1,4 +1,5 @@
-import React from 'react'
+// TODO: Figure out why importing the package name does not work.
+import { ClientEmbed } from '../../../dotcom-ui-client-embed/src'
```
nickcolley, line 2: I'm going to have another go at this but I could not figure out why referencing the package name does not work.

packages/dotcom-ui-flags/src/__test__/server/formatFlagsJSON.spec.ts
```diff
@@ -3,14 +3,8 @@ import subject from '../../server/formatFlagsJSON'
 const fixture = Object.freeze({ foo: 1, bar: false, baz: 'qux' })
 
 describe('dotcom-ui-flags/src/server/formatFlagsJSON', () => {
-  it('returns a stringified object', () => {
```
nickcolley, line 6: I've removed this since it seemed unnecessary to check the type when we do a strict assertion on an object below. Let me know what you'd prefer to do.

packages/dotcom-ui-flags/src/server/formatFlagsJSON.ts
```diff
@@ -11,5 +11,5 @@ export default function formatFlagsJSON(flags: TFlagsData = {}): string {
     }
   })
 
-  return JSON.stringify(output)
+  return output
```
nickcolley, line 14: Breaking change?

packages/dotcom-ui-flags/src/components/FlagsEmbed.ts
nickcolley: I think this was because when I added these as a dependency I just had not ran `npm install` on the directory.

packages/dotcom-ui-flags/src/server/formatFlagsJSON.ts
andygout: Hmm. Both flags and appContext already seem to be in object form (rather than a stringified object), which is surprising.

This is a screen grab from the front page:
Screenshot 2020-05-05 at 12 58 24

packages/dotcom-ui-flags/src/components/FlagsEmbed.ts
andygout: I will pull your branch and have a play.

andygout: Ah yes, that tripped me up too. If there are packages with dependencies on other packages then you need to install each of them (in order, I would imagine).

nickcolley:
Could you explain how you're testing this with the front page?

> Hmm. Both flags and appContext already seem to be in object form (rather than a stringified object), which is surprising.

This is a screen grab from the front page

`npm link`?

packages/dotcom-ui-app-context/src/client/loadAppContext.ts
```diff
@@ -1,18 +1,7 @@
+import { loadClientEmbed } from '@financial-times/dotcom-ui-client-embed/src/client'
```
nickcolley, line 1: This `src/client` business seems incorrect, I need to figure out the entry points to this package.

andygout:
> Could you explain how you're testing this with the front page?

Just checking the Dev Tools Elements tab on production ft.com, which is why it was surprising that the appContext and flags data is not stringified (as code suggests we are currently stringifying it).

examples/kitchen-sink/client/main.js
```diff
@@ -4,11 +4,15 @@ import * as layout from '@financial-times/dotcom-ui-layout'
 import * as appContext from '@financial-times/dotcom-ui-app-context'
 import * as tracking from '@financial-times/n-tracking'
 import * as ads from '@financial-times/n-ads'
+import { loadClientEmbed } from '@financial-times/dotcom-ui-client-embed'
 
 readyState.domready.then(() => {
   const flagsClient = flags.init()
   const appContextClient = appContext.init()
 
+  const customContext = loadClientEmbed('custom-context')
```
nickcolley, line 13: As you can see it's simple at the moment but does not follow the `init` convention elsewhere + the `getAll` pattern.

Should we require the user to pass their ID to `init` or automagically find clientEmbeds when `init()` is called and then allow them to use the client to specify an embed?

examples/kitchen-sink/server/controllers/home.js
```diff
@@ -64,6 +65,7 @@ module.exports = (_, response, next) => {
             </section>
           </div>
         </Layout>
+        <ClientEmbed id="custom-context" data={{ foo: true, bar: 'qux' }} />
```
nickcolley, line 68: In documentation should make it clear to put the embed at the bottom of the page to remove risk of it blocking page render.

examples/kitchen-sink/client/main.js
andygout: Would be nice if we can keep it matching the pattern.

I think including the ID as an argument for `init` will make it clear which embed is being retrieved, but means it will require an individual DOM query for each embed being retrieved (although that is what the code is currently doing, and there is only one extra embed being added to one app at the moment, so this is minimal).

nickcolley: One other alternative would be this sort of pattern:

```
@pixelandpage would appreciate your thoughts on what the interface could look like here as well if you have time :)

File: examples/kitchen-sink/client/main.js [hunk 7]
```diff
@@ -4,11 +4,15 @@ import * as layout from '@financial-times/dotcom-ui-layout'
 import * as appContext from '@financial-times/dotcom-ui-app-context'
 import * as tracking from '@financial-times/n-tracking'
 import * as ads from '@financial-times/n-ads'
+import * as clientEmbed from '@financial-times/dotcom-ui-client-embed'
 
 readyState.domready.then(() => {
   const flagsClient = flags.init()
   const appContextClient = appContext.init()
 
+  const clientEmbedClient = clientEmbed.init('custom-context')
```
nickcolley, line 13: Probably should move this 'custom-context' constant into a 'constants.js' file as per convention in this repo...

packages/dotcom-ui-app-context/src/client/AppContext.ts
```diff
@@ -1,20 +1,12 @@
...
-    this.appContext = Object.freeze(options.appContext)
-  }
-
-  get(property: string): any {
-    return this.appContext.hasOwnProperty(property) ? this.appContext[property] : undefined
-  }
-
-  getAll(): TAppContext {
-    return this.appContext
+    super(options.appContext)
```
nickcolley, line 10: Need to figure out how to extend a class and apply the types, as this currently has removed the type information.

packages/dotcom-ui-client-embed/src/client/ClientEmbedData.ts
```diff
@@ -0,0 +1,15 @@
+export default class ClientEmbedData {
```
nickcolley, line 1: Any way to add types here and then make them more specific when the class gets extended?

packages/dotcom-ui-client-embed/src/client/ClientEmbedData.ts
```diff
@@ -0,0 +1,15 @@
...
+  private data
+
+  constructor(data) {
+    this.data = Object.freeze(data)
```
nickcolley, line 5: Since the data in this is likely to be nested, should we consider doing a deeper freeze? may not be worth the extra code required in the bundle...

packages/dotcom-ui-client-embed/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './components/ClientEmbed'
+export * from './client'
```
nickcolley, line 2: I've had to add the client entrypoints to the main entrypoints which I assume is not ideal since we have a separate browser entrypoint for a reason. This was necessary to allow the client code to be shared in other packages with the TypeScript types working...

packages/dotcom-ui-app-context/src/client/AppContext.ts
nickcolley: Looks like what I wanted to do is not possible(?) so will need to figure out another way to achieve this: https://github.com/microsoft/TypeScript/issues/3402

nickcolley:
Something I noticed: 

https://github.com/Financial-Times/dotcom-page-kit/issues/809

nickcolley:
I have ruled out having a default 'id' prop as I'm not sure there's an easy way to communicate you need unique IDs if you use more than one embed.

nickcolley:
Spoke to Maggie, we're gonna go with `data-embed` to clear up the weirdness with `ClientEmbedClient` class.

nickcolley:
I've spiked putting this into next-video-page: https://github.com/Financial-Times/next-video-page/compare/spike-data-embed-2?expand=1

I still am curious if we're really solving the underlying problem. If our main worries were that app context was being passed into nTracking etc should we not have validation or a allowlist there? And allow for app context to be used for all data, which is as we've seen, what people tend to do?

andygout:
This PR description goes some way to answering that question: https://github.com/Financial-Times/dotcom-page-kit/pull/782.

I think it's preferable to enforce the schema as far upstream as possible so that Page Kit - being the originator of the `appContext` object - is able to enforce the properties it contains, ensuring that they are not outside the scope of its definition.

It's fair to say that (in the case of `next-video-page`) developers had used the `appContext` object for other data simply because it was a pre-existing way of doing what they required and therefore the most convenient solution. Even though the attribute added in `next-video-page` is called `permutiveContext`, it is just an ID and API key, which are there to allow the app run rather than serving the purpose of providing context about the app.

The `flags` object should only contain data about flags and by the same virtue the `appContext` object should only contain data about the app's context.

The concern would be that `appContext` would become a dumping ground for all sorts of values and have an inconsistent series of values across apps.

If it transpires that a value should be added to `appContext`, then this is something on which Page Kit can make the decision by making an allowance in its JSON schema validation.

nickcolley:
@andygout thanks for the clarity, let's continue. I'm still getting my head around this space. :)

nickcolley:
Worked on a different branch to clean up renaming files: https://github.com/Financial-Times/dotcom-page-kit/pull/812

Issue #799, opened 2020-04-16
Create a dotcom-ui-client-embed package for sharing data between the server and the client


pixelandpage:
Currently we have no defined method for apps to share data between the client and the server. [Some apps](https://github.com/Financial-Times/dotcom-page-kit/pull/761) have resorted to updating the `appContext` object as a way around this but we want to discourage this practice and `v0.7.0` of Page Kit introduced errors when modifying `appContext` to prevent this behaviour.

Proposal:

- Create a new dotcom-ui-client-embed package containing all of the logic required to pass data from the server to the client without modifying `appContext`.
- Refactor dotcom-middleware-app-context and dotcom-middleware-flags to use this new package rather than duplicating the same functionality in multiple places.
- Document a process for other apps to make use of the new package to pass custom properties from the client to the server.
- Incorporate this functionality into an example app.

andygout:
Case study: `next-video-page`, which currently adds this to `appContext`:

```
{
	permutiveContext: {
		publicId: process.env.PERMUTIVE_PUBLIC_ID,
		publicApikey: process.env.PERMUTIVE_PUBLIC_API_KEY
	}
}
```

nickcolley:
One thought... If someone fails the appContext schema validation we could consider pointing them to the documentation for this...

pixelandpage:
Yes, I think so 👍 At the moment I believe we serve a Typescript error but we could definitely do better than this! 

We actually have a design decisions doc about appContext which has become out-of-date; if we update that as part of this work then it might be a good resource to point to in the event of a validation error https://github.com/Financial-Times/dotcom-page-kit/blob/master/docs/design-decisions/app-context.md

PR #812, opened 2020-05-12
Add `dotcom-ui-data-embed`

nickcolley:
In order to allow users to share data between the server and client introduce a new `dotcom-ui-data-embed` package.

Some parts of this package can be reused and consumed by the `dotcom-ui-flags` and `dotcom-ui-app-context` packages.

I have [spiked this feature in the `next-video-page` application that was blocked without this feature](https://github.com/Financial-Times/next-video-page/compare/spike-data-embed-2?expand=1).

## Screenshot of kitchen sink example output
Screenshot of kitchen sink example output

Closes https://github.com/Financial-Times/dotcom-page-kit/issues/799
Closes https://github.com/Financial-Times/dotcom-page-kit/pull/761

### DIFF
--- a/examples/kitchen-sink/server/controllers/home.js
+++ b/examples/kitchen-sink/server/controllers/home.js
@@ -1,13 +1,16 @@
 const React = require('react')
 const ReactDOM = require('react-dom/server')
+const { DataEmbed } = require('@financial-times/dotcom-ui-data-embed')
 const { Shell } = require('@financial-times/dotcom-ui-shell')
 const { Layout } = require('@financial-times/dotcom-ui-layout')
 const { Slot, AdsOptionsEmbed } = require('@financial-times/n-ads')
 
+const { DATA_EMBED_ID } = require('../../constants.js')
+
 module.exports = (_, response, next) => {
   try {
     const flags = { ads: true, tracking: true }
-    const { appContext, assetLoader } = response.locals
+    const { appContext, assetLoader, embeddedData } = response.locals
     const styleBundles = [
       ...assetLoader.getStylesheetURLsFor('page-kit-layout-styles'),
       ...assetLoader.getStylesheetURLsFor('styles')
@@ -64,6 +67,7 @@ module.exports = (_, response, next) => {
             </section>
           </div>
         </Layout>
+        <DataEmbed id={DATA_EMBED_ID} data={embeddedData} />
       </Shell>
     )
 
--- a/packages/dotcom-server-app-context/README.md
+++ b/packages/dotcom-server-app-context/README.md
@@ -4,7 +4,7 @@ This package provides tools to define FT app context data and a [JSON schema] de
 
 To learn more about why this feature exists please review the [design document]. To find out which properties can be defined please refer to the [app context schema].
 
-_Please note_ that all app context properties will be appended to every tracking event sent to Spoor. For this reason it is very important not to pollute the schema.
+If you want to share application specific data with the client, consider using [@financial-times/dotcom-ui-data-embed](../dotcom-ui-data-embed).
 
 [JSON schema]: https://json-schema.org/
 [design document]: ../../docs/design-decisions/app-context.md
--- a/packages/dotcom-server-app-context/src/__test__/validate.spec.ts
+++ b/packages/dotcom-server-app-context/src/__test__/validate.spec.ts
@@ -13,7 +13,8 @@ describe('dotcom-server-app-context/src/validate', () => {
 
   it('throws an error when given an unknown property/value', () => {
     expect(() => subject('thisProperty', 'isNotInTheSchema')).toThrow(
-      'Validation error: data should NOT have additional properties, received "isNotInTheSchema"'
+      'Validation error: data should NOT have additional properties, received "isNotInTheSchema"' +
+        '\nIf you want to share application specific data with the client, consider using @financial-times/dotcom-ui-data-embed.'
     )
   })
 })
--- a/packages/dotcom-server-app-context/src/validate.ts
+++ b/packages/dotcom-server-app-context/src/validate.ts
@@ -11,6 +11,14 @@ export default function validate(field: string, value): boolean {
   if (isValid(data)) {
     return true
   } else {
-    throw Error(`Validation error: ${ajv.errorsText(isValid.errors)}, received "${value}"`)
+    let errorMessage = `Validation error: ${ajv.errorsText(isValid.errors)}, received "${value}"`
+    const hasErrorsForAdditionProperties = isValid.errors.some(
+      (error) => error.keyword === 'additionalProperties'
+    )
+    if (hasErrorsForAdditionProperties) {
+      errorMessage +=
+        '\nIf you want to share application specific data with the client, consider using @financial-times/dotcom-ui-data-embed.'
+    }
+    throw Error(errorMessage)
   }
 }
--- a/packages/dotcom-ui-data-embed/README.md
+++ b/packages/dotcom-ui-data-embed/README.md
@@ -1,0 +1,3 @@
+# @financial-times/dotcom-ui-data-embed
+
+This package provides methods for putting data into your server-side rendered pages and safely retrieving it again in the browser.
--- a/packages/dotcom-ui-data-embed/README.md
+++ b/packages/dotcom-ui-data-embed/README.md
@@ -1,3 +1,89 @@
 # @financial-times/dotcom-ui-data-embed
 
 This package provides methods for putting data into your server-side rendered pages and safely retrieving it again in the browser.
+
+## Getting started
+
+This package is compatible with Node 12+ and is distributed on npm.
+
+```sh
+npm install --save @financial-times/dotcom-ui-data-embed
+```
+
+After installing the package you can use it to embed data into your pages on the server-side. This data can then be retrieved and used in your client-side code.
+
+### Server-side integration
+
+If you are using React to render your app you can use the `DataEmbed` component to integrate the data embed data with your pages:
+
+```jsx
+import { DataEmbed } from '@financial-times/dotcom-ui-data-embed'
+const DATA_EMBED_ID = 'data-embed'
+const data = { property: 'value', secondProperty: 'second-value' }
+
+export default (props) => (
+  <html>
+    <head>
+      <meta charSet="utf-8" />
+      <title>My Amazing Website</title>
+      <DataEmbed id={DATA_EMBED_ID} data={data} />
+    </head>
+    <body>
+      ...
+    </body>
+  </html>
+)
+```
+
+Otherwise you can insert a JSON formatted string into a `<script>` element with an ID of `data-embed`.
+
+```html
+<!DOCTYPE html>
+    <html>
+    <head>
+        <meta charset="utf-8">
+        <title>My Amazing Website</title>
+        <script type="application/json" id="data-embed">
+        {"property":"value","secondProperty":"second-value"}
+        </script>
+    </head>
+    <body>
+        ...
+    </body>
+</html>
+```
+
+### Client-side integration
+
+Once you have data embedded in your page you can use the [data embed client](#client-side-api) in your client-side code. The client provides methods for safely retrieving data.
+
+```js
+import * as dataEmbed from '@financial-times/dotcom-ui-data-embed'
+
+const dataEmbedClient = dataEmbed.init({ id: 'data-embed' })
+
+if (dataEmbedClient.get('property')) {
+  ...
+}
+```
+
+## Client-side API
+
+### `init({ id }:{ id: string })`
+
+Initialises and returns a new [data embed client](#data-embed-client-api) which can be used to safely access the status of individual contexts.
+
+This method requires an `id` parameter within an options object.
+This `id` should match the `id` attribute on the embed element within the page.
+
+## Data Embed Client API
+
+### `get(property: string)`
+
+Returns the value of the requested property. If the property is not found this will return `undefined`.
+
+### `getAll()`
+
+Returns all data embed data.
+
+_Please note_ that the data returned is frozen so it cannot be modified.
--- a/packages/dotcom-ui-data-embed/README.md
+++ b/packages/dotcom-ui-data-embed/README.md
@@ -71,7 +71,7 @@ if (dataEmbedClient.get('property')) {
 
 ### `init({ id }:{ id: string })`
 
-Initialises and returns a new [data embed client](#data-embed-client-api) which can be used to safely access the status of individual contexts.
+Initialises and returns a new [data embed client](#data-embed-client-api) which can be used to safely access the data.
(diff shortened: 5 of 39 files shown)


packages/dotcom-ui-data-embed/src/client/DataEmbedStore.ts
```diff
@@ -0,0 +1,15 @@
+export default class DataEmbedStore {
+  private data
+
+  constructor(data) {
+    this.data = Object.freeze(data)
+  }
+
+  get(key: string) {
+    return this.data.hasOwnProperty(key) ? this.data[key] : undefined
```
andygout, line 9: Wouldn't `this.data[key]` return `undefined` if the key does not exist as a property on `this.data`? I.e. is the conditional check necessary here?

packages/dotcom-ui-data-embed/src/client/index.ts
```diff
@@ -0,0 +1,9 @@
+import loadDataEmbed from './loadDataEmbed'
+import DataEmbedStore from './DataEmbedStore'
+
+const init = ({ id }: { id: string }) => {
```
andygout, line 4: Not sure if it's convention, but are index files usually for just listing the modules by importing and exporting them? I.e. should this function live in its own file and be imported?

I say this when it could easily be possible that we have functions in `index` files elsewhere in this repo. Am just thinking out loud.

packages/dotcom-server-app-context/src/__test__/validate.spec.ts
```diff
@@ -13,7 +13,11 @@ describe('dotcom-server-app-context/src/validate', () => {
 
   it('throws an error when given an unknown property/value', () => {
     expect(() => subject('thisProperty', 'isNotInTheSchema')).toThrow(
-      'Validation error: data should NOT have additional properties, received "isNotInTheSchema"'
+      `
+Validation error: data should NOT have additional properties, received "isNotInTheSchema"
+
+If you want to share application specific data with the client, consider using @financial-times/dotcom-ui-data-embed.
+`.trim()
```
andygout, line 20: Backticks can be frustrating. I wonder if there's a way to do multiline without inconsistent formatting and `trim()`… 🤔

packages/dotcom-server-app-context/src/validate.ts
```diff
@@ -11,6 +11,12 @@ export default function validate(field: string, value): boolean {
...
     return true
   } else {
-    throw Error(`Validation error: ${ajv.errorsText(isValid.errors)}, received "${value}"`)
+    throw Error(
+      `
+Validation error: ${ajv.errorsText(isValid.errors)}, received "${value}"
+
+If you want to share application specific data with the client, consider using @financial-times/dotcom-ui-data-embed.
+`.trim()
+    )
```
andygout, line 20: Same backticks issue as above.

packages/dotcom-ui-data-embed/src/client/index.ts
nickcolley: This follows what `app-context` and `flags` are doing. I agree it's a bit odd though.

packages/dotcom-ui-data-embed/src/client/DataEmbedStore.ts
nickcolley: I'll admit a lot of this code is me looking for patterns in `flags` and `app-context` and consolidating the logic.

So while I didn't make the decision, I can only assume it's to avoid you being able to try to 'get' properties built into the Object.prototype e.g. `constructor` or `__proto__`.

So I'm not to sure, but perhaps could be simplified alongside `flags` and `app-context` later on?

https://masteringjs.io/tutorials/fundamentals/hasownproperty

packages/dotcom-ui-data-embed/README.md
```diff
@@ -0,0 +1,88 @@
+# @financial-times/dotcom-ui-data-embed
+
+This package provides methods for putting data into your server-side rendered pages and safely retrieving it again in the browser.
```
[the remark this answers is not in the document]
nickcolley: This is me mirroring language in `app-context` and `flags`, since the extra safety comes from the `DataEmbedStore` class rather than returning a regular object.

packages/dotcom-ui-data-embed/README.md
```diff
@@ -0,0 +1,88 @@
...
+### Client-side integration
+
+Once you have you data embedded in your page you can use the [data embed client](#client-side-api) in your client-side code. The client provides methods for safely retrieving the status of individual context properties.
+
+```js
+import * as dataEmbed from '@financial-times/dotcom-ui-data-embed'
+
+const dataEmbedClient = dataEmbed.init({ id: 'data-embed' })
+
+if (dataEmbedClient.get('property')) {
```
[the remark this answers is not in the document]
nickcolley: I think I'm mostly worried about the performance cost of shipping a deep-freeze library and I'm not too sure of the consequences of us not doing this. I suppose we want to encourage people to make copies of the data rather than relying on the ability to mutate it deeply...

Should we investigate the bundle cost of doing a deep-freeze?

packages/dotcom-server-app-context/src/__test__/validate.spec.ts
nickcolley: We could consider adding a dependency like https://www.npmjs.com/package/outdent but I felt like that was a bit much for this small usecase. What do you think?

nickcolley: We could always just do something like

```javascript
'string\n' +
'second-string` 
```

packages/dotcom-ui-data-embed/README.md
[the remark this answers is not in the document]
nickcolley: In terms of freezing flags and appContext, that should be unchanged as they don't reuse the DataEmbedStore.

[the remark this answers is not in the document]
andygout: Let's keep it just as a `.freeze` as `flags` and `appContext` currently are and worry about it if the nested properties being unfrozen ever becomes an issue.

I asked more for my curiosity.

packages/dotcom-ui-data-embed/src/client/DataEmbedStore.ts
andygout: There is always the assumption of good faith that there is a good reason behind any pre-existing code. Also happy to leave it as is for now.

packages/dotcom-ui-data-embed/src/client/index.ts
andygout: Fair enough - happy to leave as is.

examples/kitchen-sink/server/controllers/home.js
```diff
@@ -64,6 +67,7 @@ module.exports = (_, response, next) => {
             </section>
           </div>
         </Layout>
+        <DataEmbed id={DATA_EMBED_ID} data={{ foo: true, bar: 'qux' }} />
```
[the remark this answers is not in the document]
nickcolley: There can be multiple instances of `DataEmbed` if a user wants, so right now it's up to them how to get data to DataEmbed.

In theory we could replicate the middleware and server packages of app-context but that felt more than what was initially agreed in the original issue.

Then DataEmbed would just be a generic AppContext.

packages/dotcom-server-app-context/src/__test__/validate.spec.ts
andygout: Yeah, I agree with your feeling about importing a package just for this.

You could even have one really long line…

None is particularly nice so I will leave it to you to decide (which may as well be how it is currently).

packages/dotcom-ui-data-embed/src/client/DataEmbedStore.ts
nickcolley: You're right, definitely should have checked this, thanks for the prompt. :)

packages/dotcom-ui-flags/src/server/formatFlagsJSON.ts
```diff
@@ -1,6 +1,6 @@
 import { TFlagsData } from '../types'
 
-export default function formatFlagsJSON(flags: TFlagsData = {}): string {
+export default function formatFlagsJSON(flags: TFlagsData = {}): object {
```
pixelandpage, line 3: Thinking out loud, but I'm not sure this function is doing what it says anymore. If we're not formatting the list of flags as a JSON string anymore then this is filerring out the false flags rather than formatting them (which can be done much more cleanly than this if we're working with objects now!)

pixelandpage, line 3: The returned value will also be of type `TFlagsData` now, right? So `function formatFlagsJSON(flags: TFlagsData = {}): TFlagsData` will work, I think.

packages/dotcom-server-app-context/src/validate.ts
nickcolley: When I looked at this again I noticed there was an opportunity to improve the error message so we only output the additional information when they specifically encounter the error for adding additional properties. So I've managed to address the odd formatting as well at the same time...

packages/dotcom-server-app-context/README.md
```diff
@@ -4,7 +4,7 @@ This package provides tools to define FT app context data and a [JSON schema] de
 
 To learn more about why this feature exists please review the [design document]. To find out which properties can be defined please refer to the [app context schema].
 
-_Please note_ that all app context properties will be appended to every tracking event sent to Spoor. For this reason it is very important not to pollute the schema.
+If you want to share application specific data with the client, consider using [@financial-times/dotcom-ui-data-embed](../dotcom-ui-data-embed).
```
pixelandpage, line 7: 🙌

examples/kitchen-sink/server/controllers/home.js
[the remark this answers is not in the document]
pixelandpage: > In theory we could replicate the middleware and server packages 

Feels like over-engineering? 😬 We don't expect many apps to interact with this tool directly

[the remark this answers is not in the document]
nickcolley: I've pushed a commit to manually move the data into the controller: https://github.com/Financial-Times/dotcom-page-kit/pull/812/commits/3204cf6d5748a985b1f977d290cfdc943545c0de

packages/dotcom-ui-data-embed/README.md
```diff
@@ -0,0 +1,88 @@
...
+}
+```
+
+## Client-side API
+
+### `init({ id }:{ id: string })`
+
+Initialises and returns a new [data embed client](#data-embed-client-api) which can be used to safely access the status of individual contexts.
+
+Requires an `id` parameter within an options object, this `id` should match your embed `id` within the page.
```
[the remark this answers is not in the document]
nickcolley: Good point, thanks for this

packages/dotcom-ui-flags/src/server/formatFlagsJSON.ts
nickcolley: Will see if can make this better thanks

pixelandpage:
The spike in next-video-page looks good 👌 Once these changes are released we should ask the Content Innovation team to test/review the new approach in that app, then we can refine this package if needed before rolling `v0.8.x` out more widely.

examples/kitchen-sink/server/controllers/home.js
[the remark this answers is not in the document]
andygout: > There can be multiple instances of DataEmbed

So using [this usage of Page Kit](https://github.com/Financial-Times/next-subscribe/blob/85fc96193ae6446d3f76f27e6a04248445e9a0f3/server/page-kit-wrapper.js#L50-L54) as an example, it would look something like:

```
const document = React.createElement(
	Shell,
	{ ...shellProps },
	[
		React.createElement(Layout, { ...layoutProps, contents: html }),
		React.createElement(ClientEmbed, id='custom-context-foobar' data={{ foo: true, bar: 'qux' }}),
		React.createElement(ClientEmbed, id='custom-context-barfoo' data={{ foo: false, bar: 'quux' }})
	]
);
```

- This will require the document element created by the consuming app to have that many more arguments (at the moment it is quite clean with just `Shell` and `Layout`), but admittedly it is currently only one app that will be using `ClientEmbed` so is not a huge consideration.
- In keeping to the principle of least surprise, users may well expect for all the client-embedded data (flags, appContext, custom) to sit alongside one another, and that this could be achieved by passing the custom data as part of the props given to the `Shell` component - is this easily possible?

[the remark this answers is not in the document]
nickcolley: >  is this easily possible?

I think it is, but we'd want to mirror the `dotcom-server-app-context` and `dotcom-middleware-app-context` package functionality so that it is integrated in a similar way to app context which we discussed above.

So we could go ahead and flesh that functionality out, or intentionally keep this simpler and observe how people use the package before committing to something in page kit.

packages/dotcom-ui-flags/src/server/formatFlagsJSON.ts
nickcolley: Looking into this, the functionality still makes sense the main difference is I've moved when this data gets stringified to match AppContext so that they can be refactored to use DataEmbed.

examples/kitchen-sink/server/controllers/home.js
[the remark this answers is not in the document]
andygout: 👍 Okay - let's keep it as is and monitor.

packages/dotcom-ui-data-embed/README.md
```diff
@@ -0,0 +1,89 @@
...
+if (dataEmbedClient.get('property')) {
+  ...
+}
+```
+
+## Client-side API
+
+### `init({ id }:{ id: string })`
+
+Initialises and returns a new [data embed client](#data-embed-client-api) which can be used to safely access the status of individual contexts.
```
pixelandpage, line 74: Would 'properties' or 'values' work in place of 'contexts' here? I'd rather not introduce any confusion with appContext, and calling things 'context' in the past has called all manner of confusion 🙈

packages/dotcom-ui-data-embed/src/__test__/client/DataEmbedStore.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+  })
+
+  describe('.get()', () => {
+    it('returns the value of an existing data', () => {
+      expect(instance.get('foo')).toBe(1)
+      expect(instance.get('bar')).toBe(true)
+      expect(instance.get('baz')).toBe('qux')
+    })
+
+    it('returns undefined for contexts which do not exist', () => {
```
pixelandpage, line 23: See pedantic 'contexts' comment above

PR #815, opened 2020-05-14
Update Changelog for 0.8.0 feature release.

nickcolley:


changelog.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## 0.8.0 (Feature release)
+
+- Allow sharing of data between the server and client with `@financial-times/dotcom-ui-data-embed`.
```
pixelandpage, line 5: Super pedantic, but there's a convention of starting each changelog entry with the package name, e.g. `Data Embed plugin: Added a new package to provide a mechanism for passing data from server to client` and for better/worse, we don't tend to include much technical detail 🤷

changelog.md
```diff
@@ -1,5 +1,13 @@
...
+
+If you are using `@financial-times/dotcom-ui-app-context` to share data with the client you should now use the `@financial-times/dotcom-ui-data-embed` package instead.
+
+Please refer to the [`@financial-times/dotcom-ui-app-context` readme](https://github.com/Financial-Times/dotcom-page-kit/tree/master/packages/dotcom-ui-data-embed#financial-timesdotcom-ui-data-embed) for more details.
+
```
pixelandpage, line 10: It might also be worth adding the refactoring changes to the app context and flags packages? I don't think they're technically breaking changes as their interfaces haven't changed, but they're probably still significant

changelog.md
nickcolley: Will amend, I think "Allow sharing of data between the server and client" has just enough technical detail, do you agree?

changelog.md
nickcolley: I think we should only include things in the changelog which people will want to act on, so internal changes are not necessary, we could consider adding links to the pull requests for changes for people who are interested though.

PR #824, opened 2020-05-22
use an svg icon and fallback to png

JakeChampion:
png is 813 bytes and svg is 450 bytes (because we can gzip it)

### DIFF
--- a/packages/dotcom-ui-shell/src/components/DocumentHead.tsx
+++ b/packages/dotcom-ui-shell/src/components/DocumentHead.tsx
@@ -47,7 +47,7 @@ const DocumentHead = (props: TDocumentHeadProps) => (
     <link
       rel="icon"
       type="image/svg+xml"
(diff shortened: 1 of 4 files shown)


packages/dotcom-ui-shell/src/components/DocumentHead.tsx
```diff
@@ -46,12 +46,17 @@ const DocumentHead = (props: TDocumentHeadProps) => (
     {/* packaging */}
     <link
       rel="icon"
+      type="image/svg+xml"
+      href="https://www.ft.com/__origami/service/image/v2/images/raw/ftlogo-v1%3Abrand-ft-logo-square-coloured?source=update-logos"
```
nickcolley, line 50: Can we reuse the `imageServiceIconURL` function that is used in the other examples here?

JakeChampion: Sure thing :-)

PR #826, opened 2020-05-28
Use the latest version of o-fonts-assets and o-typography.

notlee:
The latest version of o-fonts-assets includes woff2 formats
for a significant reduction in filesize:
https://github.com/Financial-Times/o-fonts-assets/pull/19

 It looks like more browsers support woff2 than
`<link rel="preload">`, so it is safe to update the preloads
to the woff2 format.
https://caniuse.com/#search=woff2
https://caniuse.com/#search=preload

Bonus: When projects which depend on page kit are re-deployed
the double font download issue caused by the preload
o-fonts-assets version being different will be resolved
(for now, unless they diverge again):
https://financialtimes.slack.com/archives/C3TJ6KXEU/p1590161366213400

### DIFF
--- a/packages/dotcom-ui-base-styles/src/lib/fontFaces.ts
+++ b/packages/dotcom-ui-base-styles/src/lib/fontFaces.ts
@@ -1,8 +1,8 @@
 // NOTE: please also update ./styles.css when updating the URLs defined below. They
 // need to stay in-sync to ensure our CSS output matches the resource hints we set.
 export const fontFaceURLs = [
-  'https://www.ft.com/__origami/service/build/v2/files/o-fonts-assets@1.4.1/MetricWeb-Regular.woff',
-  'https://www.ft.com/__origami/service/build/v2/files/o-fonts-assets@1.4.1/MetricWeb-Semibold.woff',
-  'https://www.ft.com/__origami/service/build/v2/files/o-fonts-assets@1.4.1/FinancierDisplayWeb-Regular.woff',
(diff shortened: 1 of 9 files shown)


packages/dotcom-ui-base-styles/src/lib/fontFaces.ts
```diff
@@ -1,8 +1,8 @@
 // NOTE: please also update ./styles.css when updating the URLs defined below. They
 // need to stay in-sync to ensure our CSS output matches the resource hints we set.
 export const fontFaceURLs = [
```
nickcolley, line 3: Should we see both woff and woff2 files in this array now, rather than just woff2?

notlee: No because this is used for preloading, and we don't want both formats to preload. I think we're safe to just preload woff2 as it looks like browsers which support preloading also support woff2 :)

nickcolley: Nice one 👍

nickcolley:
I have tested this in `next-front-page` and `next-article` and it seems to behave as expected.

155kb (4 requests) vs 343kb (11 requests) in `next-front-page`.

Also tested in IE11 and can see the regular `.woff` files being loaded...

nickcolley:
@notlee thanks so much for this!

notlee:
thanks for suggesting we add woff2! 🎉

PR #828, opened 2020-06-04
Update changelog to include v1.0.0

andygout:


changelog.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.0.0
+
+- This is the first major stable release of Page Kit. N.B. It includes no changes since the last v0.x release.
```
nickcolley, line 5: Non-blocking, could consider removing `n.b.` as it still makes sense without it.

andygout: Pressed merge as this comment appeared.

PR #829, opened 2020-06-04
feat: add new header variant aimed at app users (ACC-396)

alexnaish:
I'm working on a card that needs to remove any outbound links in the header that may lead to content / a barrier due to iOS complications.

Apps on iOS have to pay a percentage of any subscriptions taken via the app directly to Apple so to try and mitigate the frequency of this, we want to have a header that is used on the `MyAccount` section of the website (one of the few areas of the website that is embedded on the app) that removes any links that may link to content and potentially a barrier page.

Happy to discuss options!

https://financialtimes.atlassian.net/secure/RapidBoard.jspa?rapidView=1108&projectKey=ACC&modal=detail&selectedIssue=ACC-396

Desktop Anon:
desktop-anon

Desktop Logged In:
desktop-loggedin

Mobile Logged In:
mobile-loggedin

Mobile Anon:
mobile-anon


andygout:
@pixelandpage - do you think this should be v1.1.0? It is adding new functionality/a new feature but is not a breaking change.

pixelandpage:
Really nice work @alexnaish!

Could you also update the Layout and Header package readmes to document the new option?

It might be good to explain the use-case a bit more in the PR description so that it's clear why this variant was added when developers go searching through the history in future.

PR #839, opened 2020-06-17
Update n-ui-foundations to v6.0.0

nickcolley:
Since Page kit depends on [n-topic-search](https://github.com/Financial-Times/n-topic-search/pull/34) I'd expect this to blow up, raising out of curiosity to learn more about bower dependency resolution...

Update: it did! So that's good :P Hopefully we'll see it pass once we include a new version of n-topic-search

#### References:
- [`n-ui-foundations` Migration guide v5 to v6](https://github.com/Financial-Times/n-ui-foundations#user-content-v5-to-v6).

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## 2.0.0
+
+- Shell UI Component: use an SVG icon in header and fallback to PNG, thanks @JakeChampion
+- Header UI Component: Update drawer menu accessible name for plain english alternative
+- (Breaking change) Header, Layout and Base styles UI Components: Small breaking release to update `n-ui-foundation` to version 6.
+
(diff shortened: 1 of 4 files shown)


andygout:
I included a new version of `n-topic-search` and the build passes.

nickcolley:
So should this be a breaking change then, I guess so if we've done it for n-topic-search?

andygout:
Yup, breaking change wherever we apply this upgrade so that we can be in complete control of where and when it gets consumed.

Especially so for cases like `next-article` that has `n-ui-foundations` as a dependency of several of its own dependencies, as well as a direct dependency.

nickcolley:
I'll add a CHANGELOG entry to review and we can get this merged...

PR #841, opened 2020-07-01
Add prependData option to dotcom-build-sass

andygout:
`dotcom-page-kit` uses `n-ui-foundations` in some of its packages and so was included as part of the [`n-ui-foundations` v5 to v6 migration](https://github.com/Financial-Times/n-ui-foundations#user-content-v5-to-v6) - see: https://github.com/Financial-Times/dotcom-page-kit/pull/839.

When making the final updates to a user-facing app (i.e. the endpoint of the migration), the styling of the header and footer components does not remain consistent with production:

#### Front page - local (at third commit of this PR: https://github.com/Financial-Times/next-front-page/pull/2364)
front-local

#### Front page - production
front-prod

It is caused by `o-header` using the `oGridContainer` mixin, which used to inherit the global snappy grid setting, but the global snappy setting has now been removed from `n-ui-foundations`.

Currently, these classes are added to the CSS: `o-header__container--snappy` and `o-footer__container--snappy`, but we cannot add those classes as not all apps may want this by default, i.e. it would have to configurable which would get messy as there are several locations these `--snappy` classes would need to be applied.

This approach seems the cleanest way of maintaining the consuming apps' current snappy behaviour and requires the [following change in the consuming app](https://github.com/Financial-Times/next-front-page/pull/2364/commits/a266b23393f796615955317180be39a8fe7a76ce).

The long term approach is to move away from snappy grid behaviour but it would be better if that were a planned change, rather than one we make through force of circumstance.

N.B. I have yet to address the issue of the changes fonts as I am not sure these were the intended changes made by `n-ui-foundations`.

#### References:
- [Webpack - sass-loader: prependData](https://webpack.js.org/loaders/sass-loader/#prependdata).

### DIFF
--- a/packages/dotcom-build-sass/README.md
+++ b/packages/dotcom-build-sass/README.md
@@ -69,4 +69,5 @@ The CSS loader has `@import` and `url()` resolution disabled as these should be 
 | Option            | Type     | Default | Description                                                        |
 |-------------------|----------|---------|--------------------------------------------------------------------|
 | `webpackImporter` | Boolean  | `false` | See https://github.com/webpack-contrib/sass-loader#webpackimporter |
+| `prependData`     | String   | `''`    | See https://webpack.js.org/loaders/sass-loader/#prependdata        |
 | `includePaths`    | String[] | `[]`    | See https://sass-lang.com/documentation/js-api#includepaths        |
--- a/packages/dotcom-build-sass/src/index.ts
+++ b/packages/dotcom-build-sass/src/index.ts
@@ -4,15 +4,18 @@ import type webpack from 'webpack'
 
 export type TPluginOptions = {
   includePaths?: string[]
+  prependData?: string
   webpackImporter?: boolean
 }
 
 export class PageKitSassPlugin {
   includePaths: string[]
+  prependData: string
   webpackImporter: boolean
 
-  constructor({ includePaths = [], webpackImporter }: TPluginOptions = {}) {
+  constructor({ includePaths = [], prependData = '', webpackImporter }: TPluginOptions = {}) {
     this.includePaths = includePaths
+    this.prependData = prependData
     this.webpackImporter = webpackImporter
   }
 
@@ -23,6 +26,7 @@ export class PageKitSassPlugin {
       webpackImporter: this.webpackImporter,
       // Prefer `dart-sass`.
       implementation: require('sass'),
+      prependData: this.prependData,
       sassOptions: {
         // Disable formatting so that we don't spend time pretty printing
         outputStyle: 'compressed',
--- a/packages/dotcom-build-sass/src/index.ts
+++ b/packages/dotcom-build-sass/src/index.ts
@@ -26,6 +26,9 @@ export class PageKitSassPlugin {
       webpackImporter: this.webpackImporter,
       // Prefer `dart-sass`.
       implementation: require('sass'),
+      // Prepends SCSS code before the actual entry file.
+      // Introduced to maintain snappy grid after n-ui-foundations removed it as the default.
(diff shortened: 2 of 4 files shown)


packages/dotcom-build-sass/src/index.ts
```diff
@@ -23,6 +26,7 @@ export class PageKitSassPlugin {
       webpackImporter: this.webpackImporter,
       // Prefer `dart-sass`.
       implementation: require('sass'),
+      prependData: this.prependData,
```
nickcolley, line 29: How do we feel about writing a test for this? I suppose since we're doing little and just passing through to webpack it may not be necessary?

nickcolley: Going to see how easy it is to add a test for this.

nickcolley: Looks like we could add something to the sass build example integration test.

nickcolley: sorted

packages/dotcom-build-sass/src/index.ts
```diff
@@ -4,15 +4,18 @@ import type webpack from 'webpack'
 
 export type TPluginOptions = {
   includePaths?: string[]
+  prependData?: string
```
pixelandpage, line 7: Is this a valid option for all plugins or only the sass plugin? It might be better to set up the sass plugin to extend this generic Type with the new property rather than amending it here.

packages/dotcom-build-sass/README.md
```diff
@@ -69,4 +69,5 @@ The CSS loader has `@import` and `url()` resolution disabled as these should be
 | Option            | Type     | Default | Description                                                        |
 |-------------------|----------|---------|--------------------------------------------------------------------|
 | `webpackImporter` | Boolean  | `false` | See https://github.com/webpack-contrib/sass-loader#webpackimporter |
+| `prependData`     | String   | `''`    | See https://webpack.js.org/loaders/sass-loader/#prependdata        |
```
pixelandpage, line 72: The docs suggest that the attribute name should be `additionalData`, does that matter at all?

packages/dotcom-build-sass/src/index.ts
andygout: > Is this a valid option for all plugins or only the sass plugin?
Only the sass plugin.

> It might be better to set up the sass plugin to extend this generic Type with the new property rather than amending it here.
I'm unclear on what you mean - could you please give code examples?

packages/dotcom-build-sass/README.md
andygout: This package has a dependency of `"sass-loader": "^8.0.0"`.

[Release 8.0.0](https://github.com/webpack-contrib/sass-loader/blob/master/CHANGELOG.md#user-content-800-2019-08-29) included:
> the `data` option was renamed to the `prependData` option

[Release 9.0.0](https://github.com/webpack-contrib/sass-loader/blob/master/CHANGELOG.md#user-content-900-2020-07-02), which was released just last Thursday, included:
> the `prependData` option was removed in favor the `additionalData` option, see [docs](https://github.com/webpack-contrib/sass-loader#additionaldata)

So while we are still using v8 we should use `prependData`, but once we update to the v9 then we should switch to `additionalData`.

andygout:
> This seems a little hacky. If the intention is for this to be temporary, we could think about signposting that in a comment or in the readme?

Added in commit [Add comment to explain prependData in sassLoaderOptions](https://github.com/Financial-Times/dotcom-page-kit/pull/841/commits/16028be1df971dbee1d241697eccc6e3862f4263).

PR #847, opened 2020-08-10
fix(code-splitting): split privacy modules to their own bundle

antonsamper:
#### Reference
- https://financialtimes.atlassian.net/browse/CPP-220

image

The error can be seen on the front page when you open the dev console:

```err TypeError: i.setUSPrivacy is not a function```

This is happening because something of the page is calling the method, `setUSPrivacy`, which in this case is from a module that has had its exports mangled. The issue here lies with how chunking, treeshaking, and related optimisations work within webpack. For example, the `next-front-page` module doesnt explicitly call the `setUSPrivacy` method nor does it import `n-native-ads`. So Webpack decides that this module is not being used and then shakes it out.

The fix for this is to add an [exception](https://github.com/Financial-Times/dotcom-page-kit/blob/master/packages/dotcom-build-code-splitting/src/index.ts) and create a custom bundle with the property `usedInUnknownWay: true`. This will tell Webpack not to perform optimisations which may have side effects.

I have tested this fix by updating a local copy of `dotcom-page-kit` inside the `next-front-page` `node_modules` and can confirm that it works.

***This issue could happen again in the future if a service doesn't explicitly reference a dependency and there is no exception for the module***

### CODE REVIEW

File: packages/dotcom-build-code-splitting/src/index.ts [hunk 1]
```diff
@@ -60,7 +60,15 @@ export class PageKitCodeSplittingPlugin {
       usedInUnknownWay: true
     })
 
-    // split packages used by all pages (i.e. used y Page Kit) into a shared bundle
+    // split all privacy- packages into one bundle file
+    const addPrivacyCodeSplitting = createBundleWithRegExp({
+      compiler,
+      name: 'privacy',
```
andygout, line 66: `privacy` by itself is a bit vague. Could it be renamed `privacy-components`?

nickcolley:
To make this easier to review can you please update the pull request to:

- explain what issue we've ran into
- how you know this issue will be resolved by this change
- any thoughts on the risks of this happening again in the future

carlesandres:
@antonsamper Thanks for doing this so quickly. It's massively helpful for the Ads & Privacy team.

nickcolley:
Thanks a lot for fleshing out the issue, makes it much easier to understand.

Is it possible to write a failing test for this? Then we can avoid the code splitting plugin breaking a build.

For example if we need a more real world example we could add an integration test just for this. Then if we introduce new things that'd break in this way we'll catch it in the build...

carlesandres:
@nickcolley @antonsamper FYI: at the moment we are unable to run a specific method exported from one of the `privacy` modules until this "fix" gets to production, which means we are in breach of CCPA legislation for some users.

If you think it can speed up things, would you consider writing a separate ticket/task for writing those tests?

packages/dotcom-build-code-splitting/src/index.ts
```diff
@@ -60,7 +60,15 @@ export class PageKitCodeSplittingPlugin {
       usedInUnknownWay: true
     })
 
-    // split packages used by all pages (i.e. used y Page Kit) into a shared bundle
+    // split all privacy- packages into one bundle file
+    const addPrivacyCodeSplitting = createBundleWithRegExp({
+      compiler,
+      name: 'privacy-components',
+      pattern: /[\\\/]privacy-/,
```
i-like-robots, line 67: I think this pattern is too generic and it would be sensible to prepend it with `@financial-times` to ensure it only matches the intended modules.

antonsamper: i'll update the pattern

packages/dotcom-build-code-splitting/src/__test__/index.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+          entry: {
+            scripts: path.join(__dirname, '/__fixtures__', 'entry-point.js')
+          },
+          output: {
+            filename: '[name].js',
+            path: path.join(__dirname, '/tmp')
+          },
+          plugins: [new PageKitCodeSplittingPlugin()]
+        },
+        function (err, stats) {
```
andygout, line 20:
```suggestion
        function (error, stats) {
```

In line with the [style guidelines](https://github.com/Financial-Times/dotcom-page-kit/blob/master/contribution.md#user-content-general-style) for this repo, I would suggest changing `err` here and on the following lines to `error`.

packages/dotcom-build-code-splitting/src/__test__/index.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+          if (err) {
+            throw err
+          } else if (stats.hasErrors()) {
+            throw stats.toString()
+          }
+
+          const files = stats.toJson().assets.map((x) => x.name)
```
andygout, line 27:
```suggestion
          const files = stats.toJson().assets.map(asset => asset.name)
```

Again as per the [style guidelines](https://github.com/Financial-Times/dotcom-page-kit/blob/master/contribution.md#user-content-general-style).

packages/dotcom-build-code-splitting/src/__test__/index.spec.ts
andygout, line 20: Is `stats` short for statistics here? It's unclear from its usage exactly what this is.

antonsamper: yes sort of, it contains info regarding timings and file chunks as well as any errors

https://webpack.js.org/api/node/#stats-object

PR #851, opened 2020-08-19
Webpack plugin to upload images

antonsamper:
#### Reference
- https://financialtimes.atlassian.net/browse/CPP-230

#### Notes
- Images in the applications' `client` folder (or a user defined alternative) are now uploaded automatically as part of the deployment process. References to the images in the source code are be automatically updated as part of the build step.
- Credit to @i-like-robots, @apaleslimghost and @simonplend


packages/dotcom-build-image-uploader/src/index.ts
```diff
@@ -0,0 +1,41 @@
...
+    const outputFileName = isDevMode ? '[name].[ext]' : '[name].[contenthash:12].[ext]'
+
+    compiler.options.module.rules.push({
+      test: /\.(png|jpe?g|gif|webp|ico|svg)$/,
+      use: [
+        {
+          loader: require.resolve('file-loader'),
+          options: {
+            name: (resourcePath) => {
+              const dirname = path.dirname(resourcePath)
```
i-like-robots, line 30: I think it's worth adding some comments about why this is necessary (to maintain the relative path) and maybe an explicit test too.

packages/dotcom-build-image-uploader/README.md
```diff
@@ -0,0 +1,62 @@
+# @financial-times/dotcom-build-image-uploader
```
i-like-robots, line 1: I think this package name doesn't seem right because it doesn't actually upload anything. Perhaps `dotcom-build-images` would be more appropriate?

packages/dotcom-build-image-uploader/README.md
```diff
@@ -0,0 +1,62 @@
...
+module.export = {
+  plugins: [
++    new PageKitImageUploaderPlugin()
+  ]
+}
+```
+
+## Arguments
+
+- `basePath` - Directory to look in for images. Defaults to `./client/`.
```
i-like-robots, line 34: The other `dotcom-build-` package readmes all follow the same format, I would recommend copying one of them and amending as needed for consistency and to ensure nothing is missed,

packages/dotcom-build-image-uploader/README.md
antonsamper: good shout! i think i was getting carried away with the jira ticket description

packages/dotcom-build-image-uploader/src/index.ts
apaleslimghost: "she doesn't even go here" meme

packages/dotcom-build-images/README.md
```diff
@@ -0,0 +1,56 @@
...
+`images/subdir/image.jpg` -> `images/subdir/image.jpg`
+
+The public image paths are automatically added to the generated `manifest.json`,
+meaning that they're available via the `dotcom-server-asset-loader` e.g.
+
+```
+assetLoader.getPublicURL('images/subdir/image.jpg')
+```
+
+The image files are hashed in production.
```
apaleslimghost, line 46: this is a little unclear and could do with explaining, maybe something like:

```suggestion
The image file names are hashed in production, e.g. `image.76f59deb1275.jpg`, to allow us to set longer cache times.
```

packages/dotcom-build-images/README.md
```diff
@@ -0,0 +1,56 @@
...
+	plugins: [
+		new PageKitImagesPlugin()
+	]
+}
+```
+
+## Arguments
+
+- `basePath` - Directory to look in for images. Defaults to `./client/`.
+
```
apaleslimghost, line 29: the convention for this in other plugin docs is a table, e.g. https://github.com/Financial-Times/dotcom-page-kit/tree/master/packages/dotcom-build-js#options

packages/dotcom-build-images/src/index.ts
```diff
@@ -0,0 +1,41 @@
+const path = require('path')
+const glob = require('glob')
+const MultiEntryPlugin = require('webpack/lib/MultiEntryPlugin')
+
+export class PageKitImagesPlugin {
+  basePath: string
+
+  constructor(directory = './client') {
```
apaleslimghost, line 8: other plugins use an options object instead, which i think is more readable for the consumer, and it'd be easier to add other options in the future

packages/dotcom-build-images/src/__test__/index.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+
+describe('dotcom-build-images', () => {
+  it('build images', async () => {
+    await new Promise((resolve) =>
+      webpack(
+        {
+          mode: 'none',
+          entry: {
+            // adding an example entry as webpack 4 doesnt support an empty object
+            scripts: path.join(__dirname, '/__fixtures__', 'entry-point.js')
```
apaleslimghost, line 13: i _think_ an empty array would work but i'm not sure

apaleslimghost (commented):
overall looks great, a few comments

packages/dotcom-build-images/src/__test__/index.spec.ts
antonsamper: i did try that but no luck. In Webpack 4 you cant have empty strings, arrays or objects 😞

packages/dotcom-build-image-uploader/src/index.ts
antonsamper: @i-like-robots regarding the explicit test for the relative path, this is already covered by the existing test where we check the assets that have been moved. Did you have something else in mind?

packages/dotcom-build-images/README.md
```diff
@@ -0,0 +1,58 @@
...
+		new PageKitImagesPlugin()
+	]
+}
+```
+
+## Options
+
+| Option     | Type    | Default       | Description                     |
+|------------|---------|---------------|---------------------------------|
+| `basePath` | String  | `"./client/"` | Directory to look in for images |
```
apaleslimghost, line 30: non-merge blocking: could you move this after the "Scope" section and add an example in "Scope" of using this option, so it follows the format of the other plugins?

apaleslimghost (approved):
one minor last docs comment, otherwise it's looking good!

PR #853, opened 2020-09-02
Correct subnav markup

notlee:
The class `o-header__subnav-list--subsections` does not exist.
Instead it should be `o-header__subnav-list--children`.

This will increase the padding between subnav items for PageKit
projects, which is currently too small.

Screenshot 2020-09-02 at 12 38 57


alexnaish (approved):

andygout (approved):
Ah yes, I can see that currently the gap between subnav items is much tighter than in your screenshot of the fixed state.

Screenshot 2020-09-07 at 09 31 50

PR #858, opened 2020-09-16
MAR-206 Add optimizely script to head

benliger1987:
In relation to ticket: [MAR-206](https://financialtimes.atlassian.net/browse/MAR-206)

### DIFF
--- a/packages/dotcom-ui-shell/src/components/OptimizelyHead.tsx
+++ b/packages/dotcom-ui-shell/src/components/OptimizelyHead.tsx
@@ -1,0 +1,7 @@
+import React from 'react'
+
+const OptimizelyHead = () => {
+  return <script src="https://cdn.optimizely.com/js/18203511709.js" />
+}
+
+export default OptimizelyHead
--- a/packages/dotcom-ui-shell/src/components/Shell.tsx
+++ b/packages/dotcom-ui-shell/src/components/Shell.tsx
@@ -14,6 +14,7 @@ import { FlagsEmbed, TFlagsEmbedProps } from '@financial-times/dotcom-ui-flags'
 import { Bootstrap, TBootstrapProps } from '@financial-times/dotcom-ui-bootstrap'
 import * as polyfillService from '@financial-times/dotcom-ui-polyfill-service'
 import formatAttributeNames, { TAttributeData } from '../lib/formatAttributeNames'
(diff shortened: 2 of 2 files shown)


nickramsbottom (approved):
Looks good to me!

apaleslimghost (changes requested):
your ticket says this tag should be added to the barrier pages; adding it here will add it to every user-facing app.

we also don't want to use Page Kit for short-term experiments; its predecessor was a tarpit of unmaintained, unowned stuff and we don't want to end up there again.

you've got a couple of options to add this just to the barrier pages:

- you could use the [`additionalMetadata` prop](https://github.com/Financial-Times/dotcom-page-kit/blob/master/packages/dotcom-ui-shell/README.md#additionalmetadata-reactnode) for the `Shell` component in `next-subscribe`. [here's an example in `next-front-page`](https://github.com/Financial-Times/next-front-page/blob/495018d30dd0504d95e6f3931a34caf993507f24/server/page-kit-wrapper.js#L76)
- you could possibly add the script using Google Tag Manager. i don't know much about that, you'd have to talk to the Ads & Privacy team about it.

kavanagh:
This ought to have a kill switch flag so we can turn it off in emergencies. I believe the Google Tag Manager has a similar flag passed down into it.

kavanagh (changes requested):
This ought to have a kill switch flag so we can turn it off in emergencies. I believe the Google Tag Manager has a similar flag passed down into it.

Also what @apaleslimghost about keeping experiments out of Page Kit. The other thing to try to avoid is code used for one or two apps. It's probably better if that code lives in the App until such time that we know it's part of the shell of most pages.

packages/dotcom-ui-shell/src/components/OptimizelyHead.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react'
+
+const OptimizelyHead = () => {
+  return <script src="https://cdn.optimizely.com/js/18203511709.js" />
```
nickcolley, line 4: When the document is loaded it will stop to parse scripts in the `<head>` element.

This means it's likely that this implementation will drastically slow down page load times.

benliger1987:
Please ignore this PR as I will only load the script for barrier pages. I will close soon

PR #859, opened 2020-09-21
Upgrade Node to v14

andygout:
We want to avoid the likes of `>= 12.0.0` in `package.json` files as it allow for a non-LTS (long-term support) version to be used, e.g. if v13 (non-LTS) had become available but v14 (LTS) had not yet become available. This happened on some other apps previously and required investigation to figure out the cause.


andygout:
We are sticking with Node v12 for now.

PR #860, opened 2020-09-22
Upgrade o-header to 8.5.0.

notlee:
- drawer css has been added to o-header: https://github.com/Financial-Times/o-header/releases/tag/v8.5.0
- sticky `display` was immediately reset to block so can be removed
- myft relative positioning was added for an old tooltip: https://github.com/Financial-Times/n-ui/pull/592
- sticky header is hidden for print as part of o-header: https://github.com/Financial-Times/o-header/blob/d93390313442713629da29b470b44f42e77f90d8/src/scss/features/_sticky.scss#L17

### DIFF
--- a/packages/dotcom-ui-header/src/header.scss
+++ b/packages/dotcom-ui-header/src/header.scss
@@ -1,20 +1,8 @@
-
-// Drawer
-.o-header__drawer-current-edition {
-  color: oColorsByName('black-60');
-  margin: 0.5em 0;
-}
-
-.o-header__drawer-search {
-  border-top: 2px solid oColorsByName('black-10');
-}
-
-// Sticky header
-.o-header--sticky {
-  display: none;
-}
-
 // Search
+// At the time of writing o-header duplicates search markup.
+// Once for the a experience, once for an enhanced experience.
+// Instead since we can rely on the `core` class we can use the
+// one core experience search block and reveal it if needed.
 .core [data-o-header-search] {
   display: block;
 }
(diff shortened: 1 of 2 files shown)


packages/dotcom-ui-header/src/header.scss
```diff
@@ -1,20 +1,8 @@
...
-}
-
-// Sticky header
-.o-header--sticky {
-  display: none;
-}
-
 // Search
+// At the time of writing o-header duplicates search markup.
+// Once for the a experience, once for an enhanced experience.
```
notlee, line 3:
```suggestion
// Once for a core experience, once for an enhanced experience.
```

packages/dotcom-ui-header/src/header.scss
```diff
@@ -1,20 +1,8 @@
...
+// Instead since we can rely on the `core` class we can use the
+// one core experience search block and reveal it if needed.
```
notlee, line 5:
```suggestion
// one enhanced experience search block but reveal it if needed.
```

andygout (approved):

notlee:
Thanks for the review @andygout! Unfortunately I don't have permission to release, could someone on the team release for me when they have time? :)

andygout:
@notlee - I will aim to merge this today.

My feeling is that, despite a minor upgrade being made to `o-header`, that the visual changes seen by apps that consume Page Kit will be minimal and that the Page Kit release can be a patch - would you agree?

notlee:
Thanks @andygout!  Yeah I think you're right, it should be a patch. There should be no visual changes for page kit users.

PR #868, opened 2020-10-27
Align .nvmrc with engines and CircleCi

nickcolley:


docs/design-decisions/scope.md
```diff
@@ -0,0 +1,52 @@
+# Design decisions: Scope
```
nickcolley, line 1: Unintentional files, will remove.

antonsamper (approved):

PR #869, opened 2020-10-28
CPP-341 - Upgrade storybook to version 6

nickcolley:
In the future this'll allow us to adopt newer features such as Storybook composition:
https://medium.com/storybookjs/storybook-composition-af0da9084fba

Upgrades to version 6, includes addons-essentials which is the recommended addon preset.

Followed this guide: https://medium.com/storybookjs/storybook-6-migration-guide-200346241bb5


packages/dotcom-ui-header/src/__stories__/story.tsx
```diff
@@ -15,15 +14,6 @@ import profileStoryData from './story-data/profile'
 import '../../styles.scss'
 import './demos.scss'
 
-const toggleUserStateOptions = () => boolean('Enable user nav actions', true)
```
nickcolley, line 19: The knobs version seemed to have way less duplication, I wonder if there's a better way to use Controls to avoid so much duplication?

nickcolley: Perhaps we could have all the controls as part of the default export at the top.

And disabled them like so:

https://storybook.js.org/docs/react/essentials/controls#disable-controls-for-specific-properties

package.json
```diff
@@ -18,9 +18,8 @@
     "bower:install": "npm run bower:link && bower install",
     "postinstall": "athloi exec npm i -- --no-package-lock && npm run bower:install",
     "prettier": "prettier --write '**/*.{ts,tsx,js,jsx,json}'",
-    "storybook": "npm run storybook:config && npm run storybook:start",
+    "storybook": "npm run storybook:start",
     "storybook:ci": "npm run storybook:config && npm run storybook:start:ci",
```
nickcolley, line 22: this need sorting, probably can simplify this to two scripts.

packages/dotcom-ui-header/src/__stories__/story.tsx
nickcolley: One other argument would be that these components are different enough they should have their own stories entirely maybe?

packages/dotcom-ui-header/src/__stories__/story.tsx
```diff
@@ -16,104 +14,244 @@ import profileStoryData from './story-data/profile'
...
+  showMegaNav: true,
+  showUserNavigation: true,
+  userIsLoggedIn: false,
+  currentPath: '/',
+  showLogoLink: false
+}
+DefaultHeaderWithDrawer.argTypes = {
+  showSubNavigation: {
+    control: {
+      type: 'boolean'
```
apaleslimghost, line 49: my understanding was that for simple controls like boolean values, it would be able to infer the control type, so we wouldn't need an `argType` for these, only the ones where the inferred type isn't helpful (like `currentPath` probably)? think this would remove a lot of the duplication.

nickcolley: ohhh, yeah maybe this and the disabling could do the trick, thank you.

packages/dotcom-ui-header/src/__stories__/story.tsx
nickcolley: Simplified by disabling and using inferred booleans as per Kara's suggestion...

packages/dotcom-ui-header/src/__stories__/story.tsx
nickcolley: Updated...

nickcolley:
@apaleslimghost ready for another review when you have some time, thanks for your feedback so far.

packages/dotcom-ui-footer/src/__stories__/story.tsx
```diff
@@ -1,37 +1,45 @@
...
-    'Theme',
-    {
-      dark: 'dark',
-      light: 'light'
-    },
-    'dark'
+export default {
+  title: 'FT / Footer',
+  args: { theme: 'dark' },
+  argTypes: {
```
antonsamper, line 13: Question (I've not used storybook before): 

As the components are written in typescript, do we need to specifically add `argTypes`? can we get that for free?

nickcolley: It seems to default to a text input and not infer types from TypeScript so I dont think so unfortunately.

antonsamper (approved):

PR #871, opened 2020-11-03
CPP-346 - Spike: Use npm 7

nickcolley:


makefile
```diff
@@ -3,14 +3,20 @@
 # here from next-land. this makefile should not be used for anything else.
 # instead, use per-package npm scripts
 
-clean:
+check_engine:
```
nickcolley, line 6: Make we could move this sort of check into `n-gage`?

PR #872, opened 2020-11-06
CPP-298 Wait for approval on CircleCI builds from nori and renovate PRs

emortong:
Renovate and Nori create many PRs which queue loads of builds in CircleCI blocking other PRs across the entire organisation from building, as a solution we are having PRs created by Renovate and Nori pause from building until approved. <br/><br/>This PR was created using a nori script ?<br/><br/>_Nori is a command-line application for managing changes across multiple (usually Github) repositories._

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,12 +37,15 @@ references:
   #
   # Filters
   #
-  filters_branch_build: &filters_branch_build
+  filters_only_renovate_nori: &filters_only_renovate_nori
     branches:
-      ignore:
-        - gh-pages
+      only: /(^renovate-.*|^nori\/.*)/
+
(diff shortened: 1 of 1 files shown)


emortong:
Before merging this would need to be removed, I'll just need approval to remove this required check

image

andygout (commented):
Is my understanding correct that we need to uncheck `build-test` because for Nori and Renovate branches this will not be run, and therefore such branches will be blocked because it is insisting on a process being run that never will be?

emortong:
@andygout yes. There are also required checks for the jobs `build` and `test` on their own, which are run by the `build-test` worflow and also by the `renovate-nori-build-test` workflow, so I think it is safe to remove that `build-test` required check

andygout (approved):

PR #880, opened 2020-12-30
CPP-344 call gh pages script

emortong:
DO NOT MERGE -- created PR to trigger build


emortong:
Closing since we are going with the `deploy-storybook` command instead of script approach

PR #882, opened 2021-01-05
CPP-344 Deploy Storybook to Github Pages

emortong:
- Added a deploy step to the build-test CircleCI workflow to run `deploy-storybook:ci` this command calls the storybook-deployer which builds Storybook and pushes the built files to a `gh-pages` branch.

- Added a note on the README about the Storybook deployment and where the GH_TOKEN env variable that is needed to deploy to GH pages comes from.

- The changes on the `.storybook/webpack.config.js` file are from an approach that was taken on x-dash (https://github.com/Financial-Times/x-dash/pull/561) to fix an error that was coming up when building Storybook.

- Removed `.storybook/README.md` since it was leftover from before the Storybook migration to v6.


apaleslimghost (approved):
nice! looking forward to having all this working together 👀

PR #883, opened 2021-01-19
added functionality for ifEqualsSome in dotcom-server-handlebars

gitGremlin:
A quick PR to add an additional helper to handlebars, `ifEqualsSome`
This should check to see if any condition matches the first argument.


packages/dotcom-server-handlebars/src/helpers/README.md
```diff
@@ -78,6 +78,16 @@ Example:
 
 [strictly equal]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness
 
+### ifEqualsSome
+
+Outputs the content if some of the parameters are [strictly equal].
```
emortong, line 83: Could there be a note here saying that the first parameter will be used as a control and the remaining will be compared against it? Just to be a bit more explicit.

gitGremlin: Added this for ifEquals as well as ifEqualsSome, just to make it crystal clear on the functionality of both!

emortong (approved):

examples/kitchen-sink/__test__/build.test.js
```diff
@@ -20,7 +24,6 @@ const expected = [
   'regenerator-runtime.bundle.js',
   'scripts.bundle.js',
   'shared.stable.bundle.js',
-  'shared.volatile.bundle.js',
   'styles.css',
   'webpack-runtime.bundle.js'
 ]
```
andygout, line 29: How did changes to the Handlebars helpers result in a change to the expected kitchen sink build JS manifest files?

PR #900, opened 2021-06-07
added isolatedModules field to tsconfig

noelenwenede:
This PR adds --isolatedModules with a value of `true` to tsconfig. 

Some packages in this workspace re-exports modules imported as types as regular modules and not types. 

It is difficult to track which imports should be `types` or `modules` without  ` --isolatedModules` the flag. This is the case because of the export flow which leaves the decision of the export type (type || module) to the parent file.

See https://www.typescriptlang.org/tsconfig#isolatedModules for more information.


apaleslimghost (approved):
hey @noelenwenede, great find! this looks good, thanks!

PR #903, opened 2021-07-09
Add Community & Events into the footer

ahmadmoawad:
- Add Community & Events into the footer
- Implement new Footer design

`Large screen`

Screen Shot 2021-07-09 at 3 38 25 PM

---

`Medium screen`

Screen Shot 2021-07-12 at 1 33 25 PM

---

`Small screen`

Screen Shot 2021-07-12 at 1 33 39 PM


packages/dotcom-ui-footer/bower.json
```diff
@@ -1,7 +1,10 @@
 {
   "name": "dotcom-ui-footer",
   "dependencies": {
-    "o-footer": "^7.0.0"
+    "o-footer": "9.0.0-bower"
```
ivomurrell, line 4:
```suggestion
    "o-footer": "^9.0.0-bower"
```
**nitpick:** We probably want to use a [caret range](https://docs.npmjs.com/about-semantic-versioning#using-semantic-versioning-to-specify-update-types-your-package-can-accept) so we can pick up any minor releases to `o-footer`

packages/dotcom-ui-footer/src/components/partials.tsx
```diff
@@ -63,31 +63,27 @@ const FooterContents = ({ footerData }: TFooterContentsProps) => (
         // The Next navigation API splits footer links data into "columns"
         // <https://github.com/Financial-Times/next-navigation-api/blob/HEAD/server/transforms/footer.js>
         const submenu = item.submenu.items as TNavMenuItem[][]
-
         return (
-          <div
-            key={`group-${index}`}
-            className={`o-footer__matrix-group o-footer__matrix-group--${submenu.length}`}>
+          <div key={`group-${index}`} className={`o-footer__matrix-group o-footer__matrix-group--1`}>
```
maxcbc, line 67: The className seems to have changed here, is this intentional?
It has gone from `o-footer__matrix-group o-footer__matrix-group--${submenu.length}` to `o-footer__matrix-group o-footer__matrix-group--1`

packages/dotcom-ui-footer/src/components/partials.tsx
```diff
@@ -63,31 +63,27 @@ const FooterContents = ({ footerData }: TFooterContentsProps) => (
...
             <SectionTitle label={item.label} submenu={submenu} index={index} />
             <SectionLinks submenu={submenu} index={index} />
           </div>
         )
       })}
+      <div className="o-footer__matrix-group o-footer__matrix-group--1">
+        <h3 className="o-footer__matrix-title o-footer__matrix-title--external-link">
+          <a
+            className="o-footer__matrix-link o-footer__matrix-link--more"
+            id="o-footer-section-5"
```
maxcbc, line 77: This id should be generated programmatically I think, as we don't know for sure it is number 5.
I think it should be `o-footer-section-${footerData.length}`

maxcbc (changes requested):
There are a few formatting/white-space changes that we should probably avoid as it can make the git history confusing. Also there is a couple of hardcoded numeric ids that I think should be programmatically generated (but I could be wrong).

Other than that looks good. 🙂

ivomurrell (commented):
**issue:** There seems to be some text overflow from one column to the next: 
FT Footer showing overflow from each column to the next, manifesting as text overlapping with other words
(Note the 'Today's Newspaper (ePaper)' for instance.)
**suggestion:** It looks like you can see some similar behaviour in the `o-footer` demo by resizing the window. Probably something that needs to be fixed upstream?

ahmadmoawad:
> **issue:** There seems to be some text overflow from one column to the next:
> FT Footer showing overflow from each column to the next, manifesting as text overlapping with other words
> (Note the 'Today's Newspaper (ePaper)' for instance.)
> **suggestion:** It looks like you can see some similar behaviour in the `o-footer` demo by resizing the window. Probably something that needs to be fixed upstream?

Noted

ahmadmoawad:
> **issue:** There seems to be some text overflow from one column to the next:
> FT Footer showing overflow from each column to the next, manifesting as text overlapping with other words
> (Note the 'Today's Newspaper (ePaper)' for instance.)
> **suggestion:** It looks like you can see some similar behaviour in the `o-footer` demo by resizing the window. Probably something that needs to be fixed upstream?

Hi @ivomurrell ,

The `o-footer` has been fixed and the new release now available`v9.0.1-bower`

emortong (approved):
LGTM, but I would wait for @ivomurrell 's approval as well before merging since he was reviewing it last week

packages/dotcom-ui-footer/bower.json
maxcbc: Unfortunately, as o-footer isn't pushed to bower by default any more, we need to fix the version here. Otherwise `^9.0.0-bower` resolves to `9.0.0` and caused dependency issues due to origami moving to `npm`

ivomurrell: Aaah that will be why I was having issues with storybook then! Looks to be fixed now, thanks.

ivomurrell (approved):
Footer looks good now, thanks!

maxcbc (approved):

PR #904, opened 2021-07-20
Add package-lock.json lockfiles

ivomurrell:
n-gage now supports package-lock.json files (see justification [here](https://github.com/Financial-Times/n-gage/releases/tag/v8.0.0).) Let's add lockfiles to all the packages to potentially improve caching and CI speeds.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,8 +5,6 @@ dist/
 .DS_Store
 Thumbs.db
 npm-debug.log
-package-lock.json
-yarn.lock
 .vscode
 .cache
 /**/__test__/tmp
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,0 @@
-package-lock=false
-shrinkwrap=false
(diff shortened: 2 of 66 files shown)


apaleslimghost:
ah, Athloi _hates_ `package-lock.json` unfortunately (or rather, npm hates generating `package-lock.json` in an Athloi monorepo with linked dependencies). i think we'll have to skip monorepos for now

ivomurrell:
> ah, Athloi _hates_ `package-lock.json` unfortunately (or rather, npm hates generating `package-lock.json` in an Athloi monorepo with linked dependencies).

Is this something we can fix in Athloi? After ssh'ing onto the CI box it seems that you can get it to install the packages properly if you rerun `npm install` a couple more times, so it feels like something surmountable, but I could be completely wrong about that

ivomurrell:
Yes, seems to just be a bug with npm like you said @apaleslimghost, not much we can do so I'll close the PR. Appears to no longer be an issue for me when using npm 7 so we should be able to reconsider adding lockfiles to our monorepos as part of that migration.

PR #917, opened 2021-11-18
CPP-675 Update to npm 8

ivomurrell:
We are moving to npm 8 so we are ready to migrate away from bower. There was one breaking change where `jest-puppeteer` so that the newer version of `puppeteer` we were using was within its peer dependency requirements. The major update does not seem to have affected any other behaviour. Apart from that, it was as simple as pinning a new npm version with volta (you can also update your npm globally to test the changes) and making sure CircleCI uses the right npm version too. We have added a `preinstall` script to each package's `package.json` that will make sure that npm 8 is selected when running an install.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,8 @@
     "prettier": "prettier --write '**/*.{ts,tsx,js,jsx,json}'",
     "storybook": "start-storybook -p 9001 -c .storybook",
     "storybook:ci": "start-storybook -c .storybook --ci --smoke-test",
-    "deploy-storybook:ci": "storybook-to-ghpages --ci"
+    "deploy-storybook:ci": "storybook-to-ghpages --ci",
+    "preinstall": "[ \"$INIT_CWD\" != \"$PWD\" ] || npm_config_yes=true npx check-engine"
   },
   "lint-staged": {
     "**/*.{ts,tsx,js,jsx}": [
@@ -59,6 +60,7 @@
     "@typescript-eslint/parser": "^3.0.0",
     "babel-loader": "^8.0.4",
     "bower": "^1.8.8",
+    "check-engine": "^1.10.1",
     "css-loader": "^3.0.0",
     "enzyme": "^3.10.0",
     "enzyme-adapter-react-16": "^1.15.1",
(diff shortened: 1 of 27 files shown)


serena97 (approved):

packages/dotcom-build-base/package.json
```diff
@@ -29,12 +30,16 @@
     "webpack": "^4.39.2"
   },
   "engines": {
-    "node": ">= 12.0.0"
+    "node": ">= 12.0.0",
+    "npm": "7.x || 8.x"
```
serena97, line 34: this makes sense to me, even though we're using npm 8 for the circleci build, npm version 7 is the minimum requirement for migrating components away from bower

serena97 (approved):

package.json
```diff
@@ -59,6 +60,7 @@
     "@typescript-eslint/parser": "^3.0.0",
     "babel-loader": "^8.0.4",
     "bower": "^1.8.8",
+    "check-engines": "^1.5.0",
```
serena97, line 63: **question:** where do we need this dependency `check-engines`? I noticed that in the preinstall script we are using `check-engine` with npx

serena97 (commented):
looks good to me overall, good spot on adding the preinstall script. think we'll need to remove the extra check-engines dev dependency

package.json
ivomurrell: Installing the dependency enforces the version of `check-engines` we use with npx. In addition, by default, running `npx check-engines` in the CI without the depencency defined would also fail as npm would prompt asking whether it is okay to install `check-engines` before running (and the CI wouldn't be able to respond either way without user input or a `--yes` flag included)

serena97: that sounds sensible, but when do we use `check-engines` with npx? I can only see us using `check-engine` and not `check-engines`

ivomurrell: oof – that's a good point. I'll fix that for these PRs but that was a typo in my migration script so I'll need to run a nori script to fix it in other places at some point... I guess it's running `npx` even if the package isn't installed then? I don't know how that works!!

serena97: yeah, if the package isn't installed then the latest version of `check-engine` will be installed upon running `npx check-engine`

serena97 (approved):

PR #920, opened 2021-11-22
CPP-642 migrate from circleci/* to cimg/*

serena97:
These images will be EOL from Dec 31st, and we should migrate to the modern cimg/* equivalents.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,6 +2,7 @@ version: 2.1
 
 orbs:
   node: circleci/node@4.6.0
+  browser-tools: circleci/browser-tools@1.2.3
 
 references:
 
@@ -11,7 +12,7 @@ references:
   container_config_node: &container_config_node
     working_directory: ~/project/build
     docker:
-      - image: circleci/node:12.18-browsers
+      - image: cimg/node:12.22-browsers
 
   workspace_root: &workspace_root
     ~/project
@@ -92,6 +93,7 @@ jobs:
     <<: *container_config_node
     steps:
       - *attach_workspace
+      - browser-tools/install-chrome
       - run:
           name: Check code style
           command: npm run prettier -- --list-different


.circleci/config.yml
```diff
@@ -87,6 +90,11 @@ jobs:
     <<: *container_config_node
     steps:
       - *attach_workspace
+      - browser-tools/install-chrome
+      - browser-tools/install-chromedriver
+      - run: 
+          name: Install puppeteer chromium missing dependencies
+          command: apt-get install -y libxss1
```
ivomurrell, line 97: **suggestion:** it looks like `libxss1` is already installed in the latest `12.22-browsers` image: https://github.com/CircleCI-Public/cimg-node/blob/master/12.22/browsers/Dockerfile#L50

serena97: thank you, upgrading 12.18-browsers -> 12.22-browsers fixed the problem I was facing

serena97: 🤔  I'm still baffled as to why 12.18 threw `  /home/circleci/project/build/node_modules/puppeteer/.local-chromium/linux-756035/chrome-linux/chrome: error while loading shared libraries: libXss.so.1: cannot open shared object file: No such file or directory
`
when it has libxss1 installed (here's [the dockerfile diffs](https://www.diffchecker.com/bScykl6w))

ivomurrell: Where did you get the 12.18 dockerfile from? Couldn't see it on GitHub but running `docker inspect` it doesn't look like it installs `libxss1` like the 12.22 dockerfile does

serena97: I had to fork down the repo to [generate the dockerfile](filters) for 12.22, but interesting that you can't see it has libxss1 in 12.18 through docker inspect

ivomurrell: The link isn't working but they seem to have changed the scripts they use to generate their docker images since they pushed the 12.18 image so that might explain the discrepancy

serena97: I just tested it with docker and can confirm that I see libxss1 in cimg/node:12.22-browsers but not in cimg/node:12.18-browsers

ivomurrell (approved):

PR #925, opened 2022-01-31
cpp-745 switch page kit packages to use peerdeps for origami components

serena97:


leannecornish-ft (approved):
Looks good to me! I guess this will be released as a new major version?

PR #927, opened 2022-02-07
cpp-722 widen react versions for peer deps

leannecornish-ft:
Page Kit packages currently only allow React v16 as peerDependencies. This should be widened to allow version 17 to unblock the content disco team. We could also add v18, but it's only on 18.0.0-rc.0, so not sure we'd want to do that yet?


ivomurrell (approved):

serena97 (commented):
looks good to me so far, I think we might need to add the same to dotcom-ui-base-styles as well?

PR #933, opened 2022-03-16
Added showSmartBanner to DocumentHead

phelipemaia:
Page Kit consumers will be able to decide whether they want iOs Smart Banner or not on their page.

Passing `showSmartBanner` prop as false, will not render the `apple-itunes-app` meta tag. This will be used by next-subscribe so we can hide the smart banner for users which are already using the app on subs journey.


apaleslimghost (approved):
this looks fantastic, thanks @phelipemaia! feel free to merge and release as a minor version

PR #937, opened 2022-03-21
ACQ-1536: add user subscription status as field in app context

ekocian:
As part of the following tasks https://financialtimes.atlassian.net/browse/ACQ-1393, we want to render different components based on user status in homepage, in particular, know whether the user is under the following categories: anonymous, registered, subscribed.

In that sense we want to include this information [provided by preflight](https://github.com/Financial-Times/next-preflight/blob/0157653677c5aa834bf9051393f3a6b7959f1249/server/tasks/next/ammit.js#L43) as part of the app context and make it available for the consumers.

At the moment we have already `isUserLoggedIn` which allow to know if the user is **anon** or not, so by adding this new field `isUserSubscribed` we can distinguish the other two status **registered** and **subscribed**.

### DIFF
--- a/packages/dotcom-middleware-app-context/src/index.ts
+++ b/packages/dotcom-middleware-app-context/src/index.ts
@@ -23,6 +23,7 @@ export function init(options: TMiddlewareOptions = {}) {
       // This is set by the membership session service as part of preflight
       // https://github.com/Financial-Times/next-preflight/blob/HEAD/server/tasks/membership/session.js
       isUserLoggedIn: request.get('ft-anonymous-user') === 'false',
+      isUserSubscribed: request.get('ft-user-subscription-status') === 'subscribed',
       pageKitVersion: pkg.version === '0.0.0' ? 'development' : pkg.version,
       ...options.appContext
     }
--- a/packages/dotcom-server-app-context/schema.md
+++ b/packages/dotcom-server-app-context/schema.md
@@ -78,6 +78,12 @@ If the visitor is signed in to an FT account
 
 Default: `false`
 
+## `isUserSubscribed` (boolean)
+
+If the visitor has a valid FT subscription
+
+Default: `false`
+
 ## `product` (string)
 
 The product name
--- a/packages/dotcom-server-app-context/src/schema.json
+++ b/packages/dotcom-server-app-context/src/schema.json
@@ -61,6 +61,11 @@
       "description": "If the visitor is signed in to an FT account",
       "default": false
     },
+    "isUserSubscribed": {
+      "type": "boolean",
+      "description": "If the visitor is subscribed to a FT subscription",
+      "default": false
+    },
     "product": {
       "type": "string",
       "description": "The product name",
--- a/packages/dotcom-server-app-context/src/types.d.ts
+++ b/packages/dotcom-server-app-context/src/types.d.ts
@@ -10,6 +10,7 @@ export interface TAppContext {
   conceptType?: string
   isProduction: boolean
   isUserLoggedIn?: boolean
+  isUserSubscribed?: boolean
   publishReference?: string
   [key: string]: any
 }


ludrob (approved):

jorbuedo (approved):

apaleslimghost:
hello! sorry it's taken a while for me to review this, just wanted to fully understand the context

so, currently, the subscribe button is [rendered entirely server-side](https://github.com/Financial-Times/dotcom-page-kit/blob/main/packages/dotcom-ui-header/src/components/navigation/partials.tsx#L75-L122), and the data to decide whether to enable it or not comes from the ["anonymous middleware" in n-express](https://github.com/Financial-Times/n-express/blob/77ff6c03d715a17bfca19314bee63147c452cd76/src/middleware/anon.js).

the app context is only intended for passing structured data from the server to the client, and it's only intended to be used for [event context in n-tracking](https://github.com/Financial-Times/n-tracking#options). 

unless you need this data in tracking events (and i'm pretty sure you don't, as the analytics pipeline already decorates events with data about the user's subscription level), you shouldn't need to add this property to the app context.

i would suggest:

- reading this header and setting it on `res.locals.anon` in the n-express anonymous middleware
- adding it as a property in `THeaderProps` in `dotcom-ui-header`, and using that to control how the button renders

if you've got any questions or need any more context, i'm happy to chat further about it 😊

ludrob:
> hello! sorry it's taken a while for me to review this, just wanted to fully understand the context
> 
> so, currently, the subscribe button is [rendered entirely server-side](https://github.com/Financial-Times/dotcom-page-kit/blob/main/packages/dotcom-ui-header/src/components/navigation/partials.tsx#L75-L122), and the data to decide whether to enable it or not comes from the ["anonymous middleware" in n-express](https://github.com/Financial-Times/n-express/blob/77ff6c03d715a17bfca19314bee63147c452cd76/src/middleware/anon.js).
> 
> the app context is only intended for passing structured data from the server to the client, and it's only intended to be used for [event context in n-tracking](https://github.com/Financial-Times/n-tracking#options).
> 
> unless you need this data in tracking events (and i'm pretty sure you don't, as the analytics pipeline already decorates events with data about the user's subscription level), you shouldn't need to add this property to the app context.
> 
> i would suggest:
> 
> * reading this header and setting it on `res.locals.anon` in the n-express anonymous middleware
> * adding it as a property in `THeaderProps` in `dotcom-ui-header`, and using that to control how the button renders
> 
> if you've got any questions or need any more context, i'm happy to chat further about it 😊

Makes sense, thanks @apaleslimghost!
I'm just unsure about the naming "anon" middleware , should we rename it to userStatusMiddleware or something like this?

apaleslimghost:
it's not ideal, but i think renaming the middleware would be way too large a task (it's referenced in a public option for `n-express`, `withAnonMiddleware`, which is used in dozens of repos), and tbh it already includes an `userIsLoggedIn` property which isn't strictly to do with anonymous users, so i think the cat's out of the bag on that

ekocian:
> hello! sorry it's taken a while for me to review this, just wanted to fully understand the context
> 
> so, currently, the subscribe button is [rendered entirely server-side](https://github.com/Financial-Times/dotcom-page-kit/blob/main/packages/dotcom-ui-header/src/components/navigation/partials.tsx#L75-L122), and the data to decide whether to enable it or not comes from the ["anonymous middleware" in n-express](https://github.com/Financial-Times/n-express/blob/77ff6c03d715a17bfca19314bee63147c452cd76/src/middleware/anon.js).
> 
> the app context is only intended for passing structured data from the server to the client, and it's only intended to be used for [event context in n-tracking](https://github.com/Financial-Times/n-tracking#options).
> 
> unless you need this data in tracking events (and i'm pretty sure you don't, as the analytics pipeline already decorates events with data about the user's subscription level), you shouldn't need to add this property to the app context.
> 
> i would suggest:
> 
> * reading this header and setting it on `res.locals.anon` in the n-express anonymous middleware
> * adding it as a property in `THeaderProps` in `dotcom-ui-header`, and using that to control how the button renders
> 
> if you've got any questions or need any more context, i'm happy to chat further about it 😊

Thanks! @apaleslimghost , make much more sense, I will move the logic to n-express and back to you for any doubt or clarification

PR #938, opened 2022-03-22
ACQ-1536: add subscribe message to drawer

ekocian:
**Description**
Add a new top message in drawer to remember the user to subscribe.

The implementation was made using o-message library, and it depends on the headerOptions sent by the consumer.

**Screenshots**

image
image

### DIFF
--- a/packages/dotcom-ui-header/src/components/drawer/topLevelPartials.tsx
+++ b/packages/dotcom-ui-header/src/components/drawer/topLevelPartials.tsx
@@ -24,7 +24,7 @@ const Drawer = (props: THeaderProps) => {
     >
       <div className="o-header__drawer-inner">
         <DrawerTools {...editions} />
-        {props.showSubscribeMessage && notSubscribed ? <SubscribeMessage {...props} /> : null}
+        {props.showSubscribeMessage && notSubscribed ? <SubscribeMessage /> : null}
         <Search />
         <nav className="o-header__drawer-menu o-header__drawer-menu--primary o-header__drawer-menu--border">
           {editions && <EditionsSwitcher {...editions} />}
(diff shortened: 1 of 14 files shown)


packages/dotcom-ui-header/src/components/drawer/topLevelPartials.tsx
```diff
@@ -38,6 +40,29 @@ const Drawer = (props: THeaderProps) => {
...
+          href="https://www.ft.com/products?segmentId=72adc702-369e-fc3e-1b5d-cacb5d193d80"
+          className="o-message__actions__primary"
+          data-trackable="Subscribe"
+        >
+          Subscribe
+        </a>
+      </div>
+    </div>
+  </div>
+)
```
apaleslimghost, line 64: **issue**: unfortunately, i don't think we should be including things like this in Page Kit itself, which is intended to be stable, not for features with text copy and design that we're likely to iterate on and change. 

there's some conceptual overlap with other marketing messages (i.e. delivered through `next-ammit-api` & `n-messaging-client`). the design and purpose is very similar to "top slot" messages (which also use `o-message`) such as the current "Disrupted World" message shown to anonymous users. when we were designing Page Kit, we deliberately avoided including marketing message slots to reduce the tight coupling that made its predecessor `n-ui` hard to maintain.

one option could be to add a slot to the Page Kit header here that other components can be inserted into, and build this component as a separate package that's also included by the apps that need it, so you can iterate on it and version it separately without affecting Page Kit.

(sorry to push back on this again! i hadn't seen the designs, so i thought the ticket was referring to the subscribe button in the main header, not a new component in the drawer)

ekocian:
in favor of this https://github.com/Financial-Times/dotcom-page-kit/pull/940

PR #941, opened 2022-04-08
migrate dotcom-page-kit to origami build service v3

serena97:
This PR updates relevant urls and places that generates these requests to use the v3 origami build service api, matching the urls generated in [o-fonts](https://github.com/Financial-Times/origami/blob/main/components/o-fonts/src/scss/_mixins.scss#L148) so that fonts are loaded only once. tested on examples and storybook.


serena97:
this is ready to review - just trying to fix the test-v14.19 step which is happening in other PRs too

apaleslimghost (approved):

PR #942, opened 2022-04-12
update to use npm 8.5.3 in circleci config.yml

serena97:
use the exact npm version of 8.5.3 in circleci (as per volta pin in package.json) to fix [test step not bundling o-toggle ](https://app.circleci.com/pipelines/github/Financial-Times/dotcom-page-kit/4106/workflows/166a0a4d-3d4a-479c-aa8d-9d443f2e811e/jobs/7522)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -85,7 +85,7 @@ jobs:
       - checkout
       - *restore_cache
       - node/install-npm:
-          version: "8"
+          version: "8.5.3"
       - run:
           name: Install project dependencies
           command: npm install
@@ -102,6 +102,8 @@ jobs:
     <<: *container_config_node
     steps:
       - *attach_workspace
+      - node/install-npm:
+          version: "8.5.3"
       - browser-tools/install-chrome
       - run:
           name: Check code style


.circleci/config.yml
```diff
@@ -102,6 +102,8 @@ jobs:
     <<: *container_config_node
     steps:
       - *attach_workspace
+      - node/install-npm:
+          version: "8.5.3"
```
apaleslimghost, line 106: **question**: is npm used in this job at all? why is this needed?

apaleslimghost: gotcha. does it need to be that specific version?

serena97: yes, because if it's version: "8" then it installs 8.6.0 and o-toggle somehow doesn't get bundled with this version (it also doesn't get bundled with npm 8.6.0 and node 14.19.1, but works fine locally with npm 8.5.3 and node 14.19.1 🤯 )

apaleslimghost: huh! seems weird that having a different version for the build command would cause that? happy for you to merge  but it would be good to get to the bottom of this

PR #948, opened 2022-06-13
remove duplicate parent div in NavListRightLoggedInSticky

serena97:


packages/dotcom-ui-header/src/components/sticky/partials.tsx
```diff
@@ -77,12 +77,18 @@ const NavListRightAnonSticky = (props: THeaderProps) => {
   // If user is anonymous the second list item is styled as a button
   const [signInAction, subscribeAction] = props.data['navbar-right-anon'].items
   return (
-    <div className="o-header__nav">
-      <div className="o-header__top-column o-header__top-column--right">
-        {subscribeAction && (
-          <SubscribeButton item={subscribeAction} variant="sticky" className="o-header__top-button--hide-m" />
-        )}
-        {signInAction && <SignInLink item={signInAction} variant="sticky" className="" />}
+    <div className="o-header__top-column o-header__top-column--right">
```
serena97, line 80: this parent div is still needed here - if you remove it the sticky header will look like this
Screenshot 2022-06-13 at 12 34 25 which is not what we want

ivomurrell (approved):

PR #950, opened 2022-06-30
Upgrade ft-poller from v5 to v7

alexmuller:
This is a breaking change in ft-poller which fixes a bug in that nothing listening to the `emit('error')` event results in an unhandled promise rejection warning that causes Node 16 apps to crash. It also drops support for Node 12.

See https://github.com/Financial-Times/ft-poller/pull/91 for context.

This change appears to be the first use of n-logger in Page Kit.

### DIFF
--- a/packages/dotcom-server-navigation/src/navigation.ts
+++ b/packages/dotcom-server-navigation/src/navigation.ts
@@ -1,3 +1,4 @@
+import logger from '@financial-times/n-logger'
 import Poller from 'ft-poller'
 import httpError from 'http-errors'
 import deepFreeze from 'deep-freeze'
@@ -40,6 +41,7 @@ export class Navigation {
   public options: TNavOptions
   public poller: Poller
   public initialPromise: Promise<void>
+  private menuData: TNavMenus
 
   constructor(options: TNavOptions = {}) {
     this.options = { ...defaults, ...options }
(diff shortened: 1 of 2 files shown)


joelcarr:
Link to discussion about this PR on Slack: https://financialtimes.slack.com/archives/C3TJ6KXEU/p1656683125668959?thread_ts=1656507543.908159&cid=C3TJ6KXEU

Summary of discussion:
This is a good change as we need to catch errors when getting menu data. We do not and ideally should not use `n-logger` currently on Page Kit because in theory we should be throwing the error for the consuming app to catch. However if we start throwing errors, we will cause more "H10 app crashed because uncaught error" issues. For now we will use `n-logger` and write a comment to explain the temporary solution.

packages/dotcom-server-navigation/src/navigation.ts
```diff
@@ -56,7 +58,17 @@ export class Navigation {
   async getMenusData(): Promise<TNavMenus> {
     // initialPromise does not return data but must resolve before `getData` can be called
     await this.initialPromise
-    return this.poller.getData()
+
+    try {
+      this.menuData = this.poller.getData()
+    } catch (error) {
+      // getData throws if the most recent fetch resulted in an error.
+      // In that case, continue to use the latest navigation we have available.
```
joelcarr, line 66: **request:** Please could we amend the comment to include that this is a temporary solution until there is a solution to the uncaught error issue?

joelcarr (changes requested):

alexmuller:
Sorry for the delay Joel. I've linked to a ticket where we can gather our thoughts if that works?

joelcarr (approved):
Lovely!

alexmuller:
Thanks! I don't think this will be a breaking change to the API so possibly just a minor or patch version.

PR #953, opened 2022-10-03
Ensure that all maps are called on an array

rowanmanning:
This isn't necessarily the correct solution, but wanted to get a PR open so we can discuss as a group how to solve this. I've gone with something very basic and we may want to rethink this module more.

There are cases where the Next Navigation Service will respond with `null` instead of an array. In these cases I'd prefer if we skip that part of the navigation rather than the page erroring completely which is what we saw in the incident earlier today.

**Please tear this down if there are better solutions**, but I think we need to start treating data from services that even we maintain as potentially malformed - Types alone aren't really enough for this case and I think we should be more defensive against runtime data issues.

### DIFF
--- a/packages/dotcom-ui-header/src/components/navigation/partials.tsx
+++ b/packages/dotcom-ui-header/src/components/navigation/partials.tsx
@@ -8,9 +8,13 @@ import {
   INavMeganavArticles
 } from '@financial-times/dotcom-types-navigation'
 
+const enforceArray = (value: any) => {
+  return Array.isArray(value) ? value : [];
+}
+
 const MobileNav = (props: THeaderProps) => {
   // Only display navigation on pages which are included in this menu
-  const targetUrls = props.data['navbar-simple'].items.map((item) => item.url)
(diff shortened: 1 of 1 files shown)


GlynnPhillips:
This is probably not the right place for this comment but I am going to kick it off here so it doesn't get missed and then we can move it elsewhere afterwards.

It would also be good to have some group alignment on how our API's should respond when a property is empty/missing/falsy/null.

Using this case as an example is switching `type` from `array` to `null` the correct thing for this response?

GlynnPhillips:
This is probably not the right place for this comment but I am going to kick it off here so it doesn't get missed and then we can move it elsewhere afterwards.

It would also be good to have some group alignment on how our API's should respond when a property is empty/missing/falsy/null.

Using this case as an example is switching `type` from `array` to `null` the correct thing for this response?

rowanmanning:
I'm gonna close this as @Financial-Times/cp-reliability and @Financial-Times/platforms will be working on it together in a mobbing session later this week.

PR #954, opened 2022-10-12
Remove `aria-hidden="true"` from mobile navigation

notlee:
`aria-hidden="true"` is always applied on the mobile navigation. This means assistive tech users can't navigate to markets data on mobile.
It seems unnecessary as `display: none` is set on desktop.

Screenshot 2022-10-12 at 11 51 47
Screenshot 2022-10-12 at 11 51 18


jkerr321 (approved):
Nice find, thanks Lee!

PR #955, opened 2022-10-12
CPREL-336: Fix dotcom-server-navigation by updating and improving ft-poller handling

joelcarr:
Removed need for try/catch as latest version of ft-poller no longer throws and error.


rowanmanning (approved):
Looks great

ivomurrell (approved):

PR #956, opened 2022-10-12
CPREL-337: Fix navigation type checking and add rendering fallbacks if Menu Items are missing

joelcarr:
We have aligned our navigation types to allow for `sectionlist` and `articlelist` to be `null` or `undefined`. The Origami Navigation Services allow for these type to be null but our templates did not reflect this. Our templates failed to render in https://github.com/Financial-Times/incident-reports/issues/534 when the data expected was not an array.

We have now fixed the type definition and fixed the typescript config to check for null types just for the `dotcom-ui-header` (it was turned off very early in this repo for everything and never turned back on). Turning on `strictNullChecks` means new type errors have been exposed for the package which this PR fixes.

### DIFF
--- a/packages/dotcom-ui-header/src/components/drawer/additionalPartials.tsx
+++ b/packages/dotcom-ui-header/src/components/drawer/additionalPartials.tsx
@@ -14,7 +14,7 @@ export const DrawerParentItem = ({ item, idSuffix }: TDrawerParentItemProps) => 
       <div key={item.url} className="o-header__drawer-menu-toggle-wrapper">
         <a
           className={`o-header__drawer-menu-link o-header__drawer-menu-link--${selected} o-header__drawer-menu-link--parent`}
-          href={item.url}
+          href={`${item.url}`}
           {...ariaSelected(item)}
           data-trackable={item.label}
         >
@@ -33,13 +33,13 @@ export const DrawerParentItem = ({ item, idSuffix }: TDrawerParentItemProps) => 
         id={`o-header-drawer-child-${idSuffix}`}
         data-trackable="sub-level"
       >
-        {(item.submenu.items as TNavMenuItem[]).map((item) => {
+        {(item.submenu?.items as TNavMenuItem[]).map((item) => {
           const selected = item.selected ? 'selected' : 'unselected'
           return (
             <li key={item.url} className="o-header__drawer-menu-item">
               <a
                 className={`o-header__drawer-menu-link o-header__drawer-menu-link--${selected} o-header__drawer-menu-link--child`}
-                href={item.url}
+                href={`${item.url}`}
                 data-trackable={item.label}
                 {...ariaSelected(item)}
               >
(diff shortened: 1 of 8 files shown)


rowanmanning (approved):
💪

joelcarr (commented):
We would like to add tests before merging this.

packages/dotcom-ui-header/src/components/drawer/additionalPartials.tsx
```diff
@@ -14,7 +14,7 @@ export const DrawerParentItem = ({ item, idSuffix }: TDrawerParentItemProps) =>
       <div key={item.url} className="o-header__drawer-menu-toggle-wrapper">
         <a
           className={`o-header__drawer-menu-link o-header__drawer-menu-link--${selected} o-header__drawer-menu-link--parent`}
-          href={item.url}
+          href={`${item.url}`}
```
rowanmanning, line 17: **suggestion:** I was thinking about this more and I think we need to decide between two slightly bad options. This is non-blocking but we probably want to make sure that we've made a conscious decision. After writing it up I do wonder whether option 2 is the better of the two?

## Option 1: what we have now

```jsx
<a href={`${item.url}`}>click me</a>
```

Depending on the type of `item.url` we get different results:

  * **`string`:** `<a href="https://example.com/">click me</a>`
  * **`null`:** `<a href="null">click me</a>`
  * **`undefined`:** `<a href="undefined">click me</a>`

The non-ideal thing about this is that, when these links are clicked, the user will be sent to `<current-url>/undefined`. So the user experience is maybe worse than just never showing the link.

## Option 2: empty attribute

Slightly different in that we're not casting `undefined` to a string.

```jsx
<a href={item.url ?? undefined}>click me</a>
```

Depending on the type of `item.url` we get different results:

  * **`string`:** `<a href="https://example.com/">click me</a>`
  * **`null`:** `<a>click me</a>`
  * **`undefined`:** `<a>click me</a>`

Because the attribute value is `undefined`, that attribute just doesn't get output. This might result in styling issues (I have no idea if we're styling anything based on the href attribute) but it _is_ valid HTML. From the [HTML spec](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element):

> If the [`a`](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element) element has no [`href`](https://html.spec.whatwg.org/multipage/links.html#attr-hyperlink-href) attribute, then the element [represents](https://html.spec.whatwg.org/multipage/dom.html#represents) a placeholder for where a link might otherwise have been placed, if it had been relevant, consisting of just the element's contents.

This is still maybe not an ideal user experience in the navigation, because it might be more confusing to a user to click what looks like a link and nothing happen.

## Other options

I wonder if the better option over all is to not display an anchor at all if there's no `href`, _however_ I think that's a lot more work than we have time for now – we'd have to make sure that the nav still makes sense if some elements are missing entirely. Maybe this is a later ticket to review what the behaviour should be?

rowanmanning (commented):
This is looking great, one non-blocking comment from me which I hadn't considered when we were making this change yesterday.

ivomurrell (approved):

packages/dotcom-ui-header/src/components/drawer/additionalPartials.tsx
joelcarr: We went with Option 2 as it was parity to the current wait the templates work.

PR #958, opened 2022-10-24
Add targets for workspace menu

ft-vladislav-hadzhiyski:
As discussed with Alex and Glynn in the conversation below I am submitting this PR to add `data-component` attributes to `NavListLoggedIn` and `Drawer` so we could target and add Workspace nav item with a tooltip.
image
https://financialtimes.slack.com/archives/CSH1XFM5W/p1665494436384209

### DIFF
--- a/packages/dotcom-ui-header/src/components/drawer/topLevelPartials.tsx
+++ b/packages/dotcom-ui-header/src/components/drawer/topLevelPartials.tsx
@@ -34,10 +34,7 @@ const Drawer = (props: THeaderProps) => {
         <Search />
         <nav className="o-header__drawer-menu o-header__drawer-menu--primary o-header__drawer-menu--border">
           {editions && <EditionsSwitcher {...editions} />}
-          <ul
-            data-component={props.userIsLoggedIn ? 'drawer-menu--primary__drawer-menu-list' : undefined}
-            className="o-header__drawer-menu-list"
-          >
+          <ul data-component="drawer-menu--primary__drawer-menu-list" className="o-header__drawer-menu-list">
             {primary ? <SectionPrimary {...primary} /> : null}
             {secondary ? <SectionSecondary {...secondary} /> : null}
             {tertiary ? <SectionTertiary {...tertiary} /> : null}
--- a/packages/dotcom-ui-header/src/components/drawer/topLevelPartials.tsx
+++ b/packages/dotcom-ui-header/src/components/drawer/topLevelPartials.tsx
@@ -34,7 +34,10 @@ const Drawer = (props: THeaderProps) => {
         <Search />
         <nav className="o-header__drawer-menu o-header__drawer-menu--primary o-header__drawer-menu--border">
           {editions && <EditionsSwitcher {...editions} />}
-          <ul className="o-header__drawer-menu-list">
+          <ul
+            data-component={props.userIsLoggedIn ? 'drawer-menu--primary__drawer-menu-list' : undefined}
+            className="o-header__drawer-menu-list"
+          >
             {primary ? <SectionPrimary {...primary} /> : null}
             {secondary ? <SectionSecondary {...secondary} /> : null}
             {tertiary ? <SectionTertiary {...tertiary} /> : null}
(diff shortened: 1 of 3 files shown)


packages/dotcom-ui-header/src/components/drawer/topLevelPartials.tsx
```diff
@@ -34,7 +34,10 @@ const Drawer = (props: THeaderProps) => {
         <Search />
         <nav className="o-header__drawer-menu o-header__drawer-menu--primary o-header__drawer-menu--border">
           {editions && <EditionsSwitcher {...editions} />}
-          <ul className="o-header__drawer-menu-list">
+          <ul
+            data-component={props.userIsLoggedIn ? 'drawer-menu--primary__drawer-menu-list' : undefined}
```
umbobabo, line 38: Question: why do we need the ternary here? Can't we just have the following?
```suggestion
            data-component='drawer-menu--primary__drawer-menu-list'
```

packages/dotcom-ui-header/src/components/drawer/topLevelPartials.tsx
```diff
@@ -136,7 +139,11 @@ const SectionTertiary = (props: TNavMenuItem) => (
       const divideItem = index === 0 ? 'o-header__drawer-menu-item--divide' : ''
 
       return (
-        <li key={item.url} className={`o-header__drawer-menu-item ${divideItem}`}>
+        <li
+          data-component={divideItem ? 'drawer-menu-item--divide' : undefined}
```
umbobabo, line 143: Suggestion: as per the previous question
```suggestion
          data-component='drawer-menu-item'
```

umbobabo (commented):
I believe we should change the PR description as mentioned in Slack to simplify the work of the reviewers. You can still add the link to the chat for more context.

packages/dotcom-ui-header/src/components/drawer/topLevelPartials.tsx
ft-vladislav-hadzhiyski: Here we are targeting only `o-header__drawer-menu-item--divide` items. For the Workspace menu we need only the last `o-header__drawer-menu-item--divide` item

packages/dotcom-ui-header/src/components/drawer/topLevelPartials.tsx
umbobabo: > I think the Workspace item should be visible only for logged-in users or am I wrong?
Correct, but I do would like to avoid putting any logic related to our component here. The target data attribute can be present also for anonymous users although not used. In my opinion in this case is better to have less logic in the code and have a useless data attribute on the page for anonymous users. Potentially, it could also be used in the future by someone else script.

packages/dotcom-ui-header/src/components/drawer/topLevelPartials.tsx
umbobabo: I would rather not rely on the divider but on the item themselves. What do you think about something similar to this:

```
document.querySelectorAll('.o-header__drawer-menu-item  [data-trackable="Portfolio"]')
```
With this targeting, we can clone from the Portfolio item and append before it and you will not need the modifier `--divide`

ft-vladislav-hadzhiyski: We can not rely on individual items because it's up to navigation API to return data. We can not be sure Portfolio will exist in the future as an item. The `--divide` is used to insert the Workspace link before the last divider so I don't see how we could target this in any other way

umbobabo: > We can not rely on individual items because it's up to navigation API to return data
Correct. However, this solution is meant to be temporary. At some point we want this menu to be part of the navigation and be rendered server-side only for B2B users. 
I'm happy to keep this as is 👍

packages/dotcom-ui-header/src/components/drawer/topLevelPartials.tsx
ft-vladislav-hadzhiyski: done. removed the logged in user check

ivomurrell (approved):

ivomurrell (approved):

PR #962, opened 2022-12-16
Update postcss & autoprefixer

notlee:
There is a bug in postcss autoprefixer which throws an error when `@supports selector() {}` CSS Syntax is used. This was fixed in `autoprefixer@10.2.5`.

https://github.com/postcss/autoprefixer/issues/1391

As Origami components now use `@supports selector() {}` CSS, builds for dotcom-page-kit users have begun to fail.

Upgrading autoprefixer fixes the issue. To upgrade autoprefixer postcss and postcss-loader have also
been updated.

PR #963, opened 2022-12-16
Fix TypeScript build errors

ivomurrell:
Seeing as we still aren't using a lockfile for this repository we're occasionally bumping into `@types` definitions which are doing minor version updates that are breaking TypeScript 3.x support. We need to pin the version of `@types/babel__traverse` that Jest uses in order to avoid one such compatibility issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "@types/jest": "^26.0.0",
     "@types/node": "^10.12.26",
     "@types/react": "^16.8.20",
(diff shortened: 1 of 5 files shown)


joelcarr (approved):

ivomurrell:
Closes #962

I've cherry-picked Lee's changes as they are also required to get the CI to pass

jkerr321 (approved):

PR #965, opened 2023-01-20
Put postcss-loader option in correct place

ivomurrell:
This PR fixes a bug from #964, where the `implementation` option was set for the `postcss-loader` plugin, but nested in the `postcssOptions` object when it [was meant to be outside of it](https://www.npmjs.com/package/postcss-loader#implementation). This meant that the `implementation` options was being silently ignored, and we were still running into the same resolution issues reported in the [ticket](https://financialtimes.atlassian.net/browse/CPP-1043) associated with that PR. This PR should fix that.

I was hoping that there would be a TypeScript definition for the `postcss-loader` options object so that we wouldn't run into issues like this in the future, but it seems like that doesn't exist (even though the docs are clear about what the type of each option is!) The docs are quite long and the underlying structure isn't clear unless you pay very close attention to the examples they give so we'll just have to try to be mindful of that in the future I guess.

### DIFF
--- a/packages/dotcom-build-sass/src/index.ts
+++ b/packages/dotcom-build-sass/src/index.ts
@@ -67,9 +67,9 @@ export class PageKitSassPlugin {
           // de-duplicate rule-sets which is useful if $o-silent-mode is toggled.
           // https://github.com/cssnano/cssnano
           require('cssnano')(cssnanoOptions)
-        ],
-        implementation: require('postcss')
-      }
+        ]
+      },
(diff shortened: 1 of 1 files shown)


joelcarr (approved):
Thanks for fixing my bug! 😅

ivomurrell:
I missed it in the review phase too! Had to read the [source code](https://github.com/webpack-contrib/postcss-loader/blob/v4.3.0/src/index.js#L49) before I realised we'd done it wrong...

PR #966, opened 2023-01-20
Fix failing integration tests by mirroring Navigation API change

ivomurrell:
https://github.com/Financial-Times/origami-navigation-service/pull/533 updated the `href` link for the 'Settings & Account' button in the ft.com header to be relative rather than absolute. We were testing that this element was being showing up properly when the user was signed in our integration tests, but we were selecting it by the now modified `href` attribute.

I've updated the selector to search for the new relative `href` attribute. I've also made the selector more permissive: it will find any element that _ends_ in the relative link rather than matching it exactly. It looks like Origami is planning on potentially reverting back to absolute links at some point and this more permissive selection now shouldn't break even if that change is made. Therefore, I've extended that change to all other relative links in `href` attribute selectors so that we hopefully won't have to update any of these tests in the case of switching to absolute links.

### DIFF
--- a/examples/ft-ui/__test__/anonymous-user.test.js
+++ b/examples/ft-ui/__test__/anonymous-user.test.js
@@ -6,25 +6,22 @@ describe('examples/ft-ui', () => {
 
   describe('Header link elements', () => {
     it('renders the expected anonymous user Header link elements', async () => {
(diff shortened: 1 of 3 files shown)


jkerr321 (approved):
Nice!

PR #970, opened 2023-02-22
CPP-1249: Replace athloi with npm workspaces

joelcarr:
Replace Athloi with using npm workspaces. Athloi is no longer a viable solution and now npm workspaces meets all of our needs natively. This PR replaces all the build, test, and publish workflows to use npm workspaces. This will massively improve the ability to maintain this repository as npm workspaces is able to run `npm install` with no dependency changes in **17 seconds** compared to when using Athloi (via the postinstall script) doing it in  **2 minutes 37 seconds (!!)**.


package.json
```diff
@@ -6,8 +6,8 @@
   "scripts": {
     "test": "jest",
     "test:examples": "athloi run test --filter 'example-*'",
-    "build": "athloi run build --filter 'dotcom-*' --concurrency 10 --preserve-order",
-    "build:examples": "athloi run build --filter 'example-*' --concurrency 10",
+    "build": "tsc --build --verbose tsconfig.build.json && npm run build:browser --workspaces --if-present",
```
joelcarr, line 9: I aimed to have a script to build all the typescript files for both our node and browser needs but wasn't able to figure that out. This builds all typescript files by default for node then we have a task that runs in each package with a `build:browser` script defined.

jkerr321: **praise:** Nice! It's great to read your tech lead publicly say "I wasn't able to figure that out" and it's no big deal 😁

package.json
```diff
@@ -6,8 +6,8 @@
...
+    "build:examples": "npm run build --workspace=examples/ --if-present",
     "checktypes": "tsc --noEmit",
```
jkerr321: **question:** I don't think we need line 11 (`checktypes`) now the `pre-commit` hook has been changed on line 91?

packages/dotcom-build-base/package.json
```diff
@@ -6,12 +6,9 @@
   "types": "src/index.ts",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
-    "tsc": "../../node_modules/.bin/tsc --incremental",
     "clean": "npm run clean:dist && npm run clean:node_modules",
     "clean:dist": "rm -rf dist",
     "clean:node_modules": "rm -rf node_modules",
-    "build": "npm run build:node",
-    "build:node": "npm run tsc -- --module commonjs --outDir ./dist/node",
```
jkerr321, line 14: **question:** you have fully removed, rather than edited, the `build:node` and `build` scripts back in here - was that intentional?

joelcarr: It gets removed in this commit:  https://github.com/Financial-Times/dotcom-page-kit/pull/970/commits/61621ac88625245ee0f622d95084b1a67035f50d

package.json
jkerr321: Ahh I see you have done this in a later commit! 👍

packages/dotcom-build-base/package.json
jkerr321: I think your response is not the answer to this question? 😊

joelcarr: Oops! I will add it back in to be like the other packages.

jkerr321: Great, 👍 from me once that's done!

joelcarr: Done!

packages/dotcom-build-code-splitting/package.json
```diff
@@ -6,12 +6,11 @@
...
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
-    "tsc": "../../node_modules/.bin/tsc --incremental",
     "clean": "npm run clean:dist && npm run clean:node_modules",
     "clean:dist": "rm -rf dist",
     "clean:node_modules": "rm -rf node_modules",
-    "build": "npm run build:node",
-    "build:node": "npm run tsc -- --module commonjs --outDir ./dist/node",
+    "build:node": "tsc",
+    "build": "node run build:node",
```
ivomurrell, line 13: **issue**: I don't think this script should have changed – `node run` isn't a valid command and results in an error when you run `npm run build:node` as it tries to load a module called `run`. Looks like this is present in all the `package.json`'s in the `packages` subdirectory.

joelcarr: Looks like my find and replace skills have failed me. Fixing it now!

joelcarr: Fixed 👍

.circleci/config.yml
```diff
@@ -128,12 +136,25 @@ jobs:
...
       - run:
           name: Bump version
-          command: npx athloi version ${CIRCLE_TAG} --filter "dotcom-*"
+          command: npm version ${CIRCLE_TAG} --workspace=packages/
       - run:
           name: NPM publish
-          command: |
-            export DIST_TAG=$(node scripts/get-npm-dist-tag.js $CIRCLE_TAG)
-            if [[ $DIST_TAG ]]; then npx athloi publish -- --access=public --tag=${DIST_TAG}; fi
+          command: npm publish --access=public --tag=latest
```
ivomurrell, line 142: **thought**: the only downside to this approach is that if we do a patch release for an older breaking version then that release will be tagged as `latest`. but I don't think that it's a big deal and we get the benefits of clearer configuration with more logic staying in the CircleCI configuration file, thank you!

ivomurrell (approved):
This is great, thanks for taking the time to do this and ease the pain of other developers working on the Page Kit project!!

.circleci/config.yml
joelcarr: We are losing that functionality however it would be something we rarely do and our aim with Page Kit is that everyone is using the same version.

rowanmanning:
Ooh nice work 🎉

PR #973, opened 2023-03-01
CPP-1317: fix missing files to publish in each package

joelcarr:
Before the migration to npm workspaces for publishing, we used to ignore a select group of files. Now we define the files to publish, we need to add the missing files that have not been defined but are required.

### DIFF
--- a/packages/dotcom-build-base/package.json
+++ b/packages/dotcom-build-base/package.json
@@ -33,7 +33,7 @@
     "node": ">= 14.0.0",
     "npm": "7.x || 8.x"
   },
-  "files": ["dist/"],
+  "files": ["dist/", "src/"],
   "repository": {
     "type": "git",
     "repository": "https://github.com/Financial-Times/dotcom-page-kit.git",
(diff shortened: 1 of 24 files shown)


jkerr321 (approved):

packages/dotcom-build-base/package.json
```diff
@@ -33,7 +33,7 @@
     "node": ">= 14.0.0",
     "npm": "7.x || 8.x"
   },
-  "files": ["dist/"],
+  "files": ["dist/", "src/"],
```
ivomurrell, line 36: **question**: why does `src` need to be included here? it just includes a typescript file which is transpiled into the `dist` directory

joelcarr: This was to replicate what was published when using Athloi. We were previously using `.npmignore` to ignore all files we didn't want which were related to tests and `node_modules` which meant `src` was published. For this, we have made the published packages at parity with the previous publish workflow but we should do work in the future to evaluate what we actually need to publish for each package.

PR #978, opened 2023-06-08
CPP-1486: Add support for Node 18

joelcarr:
Migrate and add support for Node 18 to Page Kit packages.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,6 @@
 version: 2.1
 
 orbs:
-  node: circleci/node@4.9.0
   browser-tools: circleci/browser-tools@1.2.4
 
 references:
@@ -16,7 +15,7 @@ references:
       - image: cimg/node:<< parameters.node-version >>-browsers
     parameters:
       node-version:
-        default: "16.14"
+        default: "16.20"
         type: string
 
   workspace_root: &workspace_root ~/project
@@ -92,8 +91,6 @@ jobs:
     steps:
       - checkout
       - *restore_cache
-      - node/install-npm:
-          version: "8.5.3"
       - run:
           name: Install project dependencies
           command: npm install
@@ -110,8 +107,6 @@ jobs:
     <<: *container_config_node
     steps:
       - *attach_workspace
-      - node/install-npm:
-          version: "8.5.3"
       - browser-tools/install-chrome
       - run:
           name: Check code style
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -1,24 +1,23 @@
 version: 2.1
 
 orbs:
-  node: circleci/node@4.9.0
   browser-tools: circleci/browser-tools@1.2.4
 
+executors:
+  node_lts:
```
ivomurrell, line 7: **nitpick**: this name won't be true in [less than 5 months](https://nodejs.dev/en/about/releases/) 😄 maybe something more generic like `node` or something more static like `node18_16` would be more future proof?

.circleci/config.yml
```diff
@@ -1,24 +1,23 @@
...
+    docker:
+      - image: cimg/node:16.20-browsers
+
 references:
 
   #
   # Workspace
   #
-  container_config_node:
-    &container_config_node
```
ivomurrell, line 13: **question**: was there a reason you stopped using the anchor? it's created some additional copy/pasted code

ivomurrell (approved):
**praise**: thank you for sorting this out, the webpack incompatibilities have made this into a tricky project to migrate!

.circleci/config.yml
joelcarr: My thinking was to make it clear that our primary version is the LTS version. However this comment has made me think - maybe we should have a more generic name like "node_present", "node_past", and "node_future" (🎄 👩‍🎤 👨‍🎤 🧑‍🎤 joke intended) which makes it clear of what the current supported version is and means that updating the minor version only has to be done on the image. What do you think?

.circleci/config.yml
joelcarr: I got ride of it initially with my lack of understanding that jobs required an executor (I was defining just in the workflow) so it looked like this anchor was only defining the `working_directory` but that isn't the case anymore. I will refactor it back in.

.circleci/config.yml
ivomurrell: I worry that those names are still too dependent on current state of the Node release schedule and will go out of date and be misleading in a few months (well `node_present` and `node_future` would anyway, the past will always be in the past, thankfully)

joelcarr: This is me trying to read the future 🔮 So making a figurative note of it for if we come across it in the future 👍

ivomurrell (approved):
🎉

PR #992, opened 2023-09-08
CPREL-711: Remove default metadata description because it causes errors in Semrush

philsegal:
From Semrush: "Duplicate meta descriptions on different pages mean a lost opportunity to use more relevant keywords. Also, duplicate meta descriptions make it difficult for search engines and users to differentiate between different webpages. It is better to have no meta description at all than to have a duplicate one."

https://developers.google.com/search/docs/appearance/title-link?visit_id=638302770161667479-4233470687&rd=1


joelcarr (commented):
Please could you add a description for the PR which informs what the change is and why it is needed?

PR #996, opened 2023-09-19
fix: flyout menu behavior

ibrahim-mamdouh-ft:
In this PR I have resolved the UI issues mentioned in this ticket [JIRA ELES-817](https://financialtimes.atlassian.net/browse/ELES-817)

There are no visual changes but the behavior is updated to be more user friendly

### DIFF
--- a/packages/dotcom-ui-header/src/enhanced-search/customList.js
+++ b/packages/dotcom-ui-header/src/enhanced-search/customList.js
@@ -13,6 +13,7 @@ class CustomSuggestionList extends BaseRenderer {
 
   renderSuggestionChip = (term) => {
     return `<a
+        tabindex="0"
         data-trackable="link"
         data-suggestion-id="${term}"
         href="${this.enhancedSearchUrl}${term}"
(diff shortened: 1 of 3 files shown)


packages/dotcom-ui-header/src/enhanced-search/customList.js
```diff
@@ -65,7 +82,7 @@ class CustomSuggestionList extends BaseRenderer {
       suggestions.push({
         linkClassName: 'enhanced-search__target--news',
         trackable: 'enhanced-search-news',
-        suggestions: this.state.suggestions.concepts.slice(0, DISPLAY_ITEMS).map((suggestion) => ({
+        suggestions: this.state.suggestions.concepts?.slice(0, DISPLAY_ITEMS).map((suggestion) => ({
```
egargan-ft, line 85: If `suggestions.concepts` is undefined, this whole expression will return undefined, is that OK? Are we null-checking this `suggestions` property when we read it?

ibrahim-mamdouh-ft: I have handled the potential undefined in renderSuggestionGroup

egargan-ft (approved):

PR #1000, opened 2023-09-22
fix: address latest QA findings

ibrahim-mamdouh-ft:
In this PR I have addressed some of the latest QA findings [here](https://financialtimes.enterprise.slack.com/files/U04B8H5442F/F05TE3GJNDT/21_sep_enhanced_search_qa_session.pdf?origin_team=T025C95MN&origin_channel=Vall_threads)

1. "Tip text shouldn’t be bold, the text after “Search tip: ...”, as per designs"
Screenshot 2023-09-22 at 1 52 55 PM

2. "Error message should read “failed to load topics and securities results” (The styles will be applied normally in other repos)"
Screenshot 2023-09-22 at 1 56 18 PM

3. "Topics column is still taking up space when empty"
Screenshot 2023-09-22 at 1 54 42 PM

4. "Can we add tracking to errors?" (Firing a custom event 'flyout-load-topics-error') when error happens
5. "Don’t show search bar on sticky header" I have set the aria-hidden property to "true" for sticky header to hide it

### DIFF
--- a/packages/dotcom-ui-header/src/enhanced-search/enhancedSearch.js
+++ b/packages/dotcom-ui-header/src/enhanced-search/enhancedSearch.js
@@ -12,6 +12,8 @@ class EnhancedSearch extends TopicSearch {
           searchTerm: this.searchEl.value,
           suggestions: {}
         })
+        // const detail = { category: 'search', action: `Error: ${error.message}` }
+        // document.body.dispatchEvent(new CustomEvent('flyout-load-topics-error', { detail, bubbles: true }))
       }
     })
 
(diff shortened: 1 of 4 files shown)


packages/dotcom-ui-header/src/enhanced-search/enhancedSearch.js
```diff
@@ -12,6 +12,8 @@ class EnhancedSearch extends TopicSearch {
           searchTerm: this.searchEl.value,
           suggestions: {}
         })
+        const detail = { category: 'search', action: `Error: ${error.message}` }
```
egargan-ft, line 15: Let's save this for a separate PR I think, the contents of this payload are pretty standardised, e.g. i think we have to choose from a standard set of values for `category` and `action`!

It might take some time to figure this out, so let's not bog down the other fixes while we figure this one out.

egargan-ft (changes requested):

egargan-ft (approved):

PR #1005, opened 2023-10-05
Remove enzyme

benwallman:
This PR removes enzyme and moves us over to react-testing library.

Enzyme has been deprecated, and it seems react-testing library is the new leader in testing React applications.


packages/dotcom-ui-layout/src/components/__test__/Layout.spec.tsx
```diff
@@ -1,227 +1,233 @@
...
-  })
+    const { container } = render(<Subject />)
+
+    const skipToNavigationLink = container.querySelector('a[href="#site-navigation"]')
+    const skipToContentLink = container.querySelector('a[href="#site-content"]')
+    const skipToFooterLink = container.querySelector('a[href="#site-footer"]')
 
-  it('renders the header and drawer components', () => {
-    expect(result.find(Header)).toExist()
-    expect(result.find(Drawer)).toExist()
```
joelcarr, line 23: **question:** What I like about the previous approach is that we could also query what properties were passed to the mock. The new method we lose that as we have completely mocked the component. Are we able to keep the previous method of shallow rendering the child components without mocking them completely?

joelcarr (commented):
Looks good! I have left a comment about one of the tests but as tests are passing and they are all translated, I am happy to approve this after we discuss the question in the comment.

packages/dotcom-ui-layout/src/components/__test__/Layout.spec.tsx
benwallman: With mocks in react testing library it's still possible to check the props passed to the children! https://stackoverflow.com/a/58634377

Secondly, it seems shallow rendering as a concept has died out with enzyme. I can't find it in the new react docs, just the legacy ones. I'd love to be proven wrong on this one however, if you know any more on this topic. A good article for why shallow is flawed if you're interested: https://kentcdodds.com/blog/why-i-never-use-shallow-rendering

contribution.md
```diff
@@ -109,17 +109,13 @@ We use [Jest] as the test runner and assertion library for all code. To run test
...
 
-Components are rendered in tests using [Enzyme] with specific assertions provided by the [enzyme-matchers] library for Jest.
-
 Code style is enforced with Prettier and non-compliant code should be automatically formatted when committing.
 
 We have implemented [ESLint] to statically analyse code for problems.
 
 [Jest]: https://jestjs.io/
 [Supertest]: https://github.com/visionmedia/supertest
 [Puppeteer]: https://github.com/smooth-code/jest-puppeteer
```
joelcarr, line 118: **suggestion(non-blocking):** We should add a link to React Testing Library https://testing-library.com/docs/react-testing-library/intro

joelcarr (approved):
This is great! Thanks so much for bringing the new standard to our tests. The tests match parity with the previous tests.

**thought:** Now we are using `@testing-library/react` it would be great to use the contextual accessible query methods but this should be a future task.

contribution.md
benwallman: Will do, but will just add it on to my next PR to keep this moving!

PR #1009, opened 2023-10-11
WIP Dont Merge - Adsdev 1491

andrewgeorgiou1981:
This PR relates to the CMP (Consent Management Platform) integration.
We are integrating with a CMP vendor named Sourcepoint and one of the requirements is that our 'Manage Cookies' links will open the CMP popup modal where a user can manage their consent choiices.

It has been decided that we should put this functionality behind a feature-flag (`adsDisableInternalCMP`). 

This PR updates the client-side portion of the `dotcom-ui-footer` package to do the following:
 * import the `dotcom-ui-flags` package so that we can make use of feature flags 
 * imports a single function from the new `privacy-utils` package which runs client-side and updates the `Manage Cookies` link element, adding an onclick handler which calls the CMP function for loading the Consent Popup Modal.

### DIFF
--- a/packages/dotcom-ui-footer/browser.js
+++ b/packages/dotcom-ui-footer/browser.js
@@ -1,4 +1,9 @@
 import Footer from '@financial-times/o-footer'
+import * as flags from '@financial-times/dotcom-ui-flags'
+import { updateFooterLinkCMP } from '@financial-times/privacy-utils'
+
+const flagsClient = flags.init()
+
 
 /**
  * @typedef FooterOptions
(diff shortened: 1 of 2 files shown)


joelcarr (commented):
I know this is a WIP but it caught my eye. Wanted to let you know of a design change requirement before you go further with the implementation.

andrewgeorgiou1981:
Closing this PR as we will go with an alternative approach. We will not be importing the privacy package into the footer package, nor we will need to import the flag-client into the pagekit packages.

PR #1014, opened 2023-11-22
Example with nextjs

benwallman:
Hey,

We've got a few examples working in dotcom-page-kit, e.g. handlebars

Having an example of NextJS working with dotcom-page-kit should show firstly how to set up a new project which uses NextJS, as well as demonstrate the versatility of dotcom-page-kits components (e.g. Layout)


joelcarr (changes requested):
While this may be possible, this is not something we are ready to support yet. The examples we have are for ways we expect Page Kit to be used. We do not expect NextJS to be used for Customer Products applications yet and will require further exploration to ensure our tooling and monitoring is compatible.

joelcarr:
Thanks for adding the note to the README however I do not believe this is enough to detach responsibility of it from us. I am going to close this PR. We may formally support Javascript framesworks like NextJS later on but we are not ready to do this now.

PR #1017, opened 2023-12-12
Account and myFT entry points test

GlynnPhillips:
### Description

The Retention team aim to run an AB test with the hope of improving the confusion regarding entry points to myFT and Settings & Account.

The changes include

1. Swapping the position of myFT and Settings & Account entry points, so Setting & Account is at the top and MyFT is below
2. Changing the name of "Settings & Account" to "My Account"
3. Adding a "user" icon next to the My Account and Sign in links
4. Adding "Feed" to the myFT link

 These changes are related to user research that discovered
 
 1. Users think the myFT logo is just branding and not a link
 2. Settings and Account was not obvious place to go as the main hub for your Account
 3. My Account was a more common term for that hub
 4. The user icon help to distinguish this as your space
 
 These changes have also been couples with other changes else where including
 
 1. Adding sign posting to myFT features from within Settings & Account (My Account) so users can still find their way to those features from a central hub
 2. Removing the login barrier from Settings & Account (My Account) landing page so that sign posting is visible without being prompted to login again
 
 ### Screenshots
 
 **Anonymous user - test control**

Screenshot of the FT header as seen by an anonymous user in the test control
 
  **Anonymous user - test variant**  
  
Screenshot of the FT header as seen by an anonymous user in the test variant

  **Signed in user - test control**
  
  
Screenshot of the FT header as seen by an registered user in the test control

  **Signed in user - test variant**
  
Screenshot of the FT header as seen by an regustered user in the test variant

**Mobile view**
This is the same regardless if you are logged in or not

Screenshot of the FT header as seen by a user in the test variant

### Interface changes

A new property has been added to the header components interface, that property is called `accountEntryTest` and it can be a `boolean` or `undefined` to allow for apps where this change doesn't need to be rolled out. That new property is temporary and will be cleaned up based on the outcome of the test.

Most (probably all) apps this is relevant to use the `dotcom-ui-layout` component to render pages so the new property will be passed into `dotcom-ui-layout` in the `headerOptions` object. For example in next-profile it will be added [here](https://github.com/Financial-Times/next-profile/blob/main/server/controllers/concerns/page-kit.js#L151)

### Ticket

https://financialtimes.atlassian.net/browse/ACC-2757

### Discussion points

As the interface change is optional and the output is unchanged without the new property then I would propose this change is released as a new minor, does anyone disagree with that?

### DIFF
--- a/packages/dotcom-ui-header/src/header.scss
+++ b/packages/dotcom-ui-header/src/header.scss
@@ -34,15 +34,26 @@
   @include oTypographySans(0);
 }
 
+
 .ft-header__top-link--myaccount span {
   vertical-align: middle;
-
+  // Hide the myaccount/sign in text on smaller screens leaving the icon only
   @include oGridRespondTo($until: 'M') {
     @include oNormaliseVisuallyHidden;
   }
 }
 
-.ft-header__top-link--myaccount:before {
+// Override the hover styles so the underline
+// Is only under the text and not the icon
+// And is closer to the text
+.ft-header__top-link--myaccount::after {
+  width: calc(100% - 32px);
+  left: unset;
+  right: 0;
+  bottom: 8px;
+}
+
+.ft-header__top-link--myaccount::before {
   content: '';
   display: block;
   @include oIconsContent(
--- a/packages/dotcom-ui-header/src/header.scss
+++ b/packages/dotcom-ui-header/src/header.scss
@@ -3,7 +3,7 @@
 // Once for a core experience, once for an enhanced experience.
 // The enhanced version is hidden until toggled with JavaScript.
 // Instead since we can rely on the `core` class we can use the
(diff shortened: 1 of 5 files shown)


packages/dotcom-ui-header/src/header.scss
```diff
@@ -27,3 +27,40 @@
 .n-typeahead {
   display: none;
 }
+
+// The styles below are part of an AB test
```
GlynnPhillips, line 31: **Note:** These styles are not designed to be perfect for a future after the test. If the test is sucessfull then the styles should correctly designed within o-header.

joelcarr (approved):
Thanks for implementing the changes we discussed offline! We discussed making the opt-in property clear that it is experimental and will likely be removed to hopefully remove any accidental use thinking its a supported feature.

PR #1018, opened 2023-12-18
fix: Remove unnecessary peer dependencies for code splitting

GlynnPhillips:
_**Theres no big urgency to this so it can wait until the new year if we think theres any risk to it**_

### Description

These peer dependencies were added in https://github.com/Financial-Times/dotcom-page-kit/pull/1012

However I don't believe they are needed as peer dependencies at all as the code splitting package doesn't use them.

It only references them so if they are used in a consuming app the code can be split into their own bundle as they are commonly used.

The issue with including these as a peer dependencies is that it forces apps to include these dependencies at a set version regardless, therefore potentially increasing the work to upgrade an app to the latest page-kit version

Maybe I am missing something here about needing to ensure the bundle we create is based on a certain version? but I cannot find any evidence of this being the case and this package work prior to these being included as a peer dependency.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2594,180 +2594,6 @@
         "mathsass": "0.10.1"
       }
     },
-    "node_modules/@financial-times/n-feedback": {
-      "version": "9.0.4",
-      "resolved": "https://registry.npmjs.org/@financial-times/n-feedback/-/n-feedback-9.0.4.tgz",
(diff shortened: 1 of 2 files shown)


GlynnPhillips:
@apaleslimghost - I can see you commented on the need for this at the time as well ([here](https://github.com/Financial-Times/dotcom-page-kit/pull/1012#discussion_r1380450197)) but I think that comment was taken as a request to add the missing ones rather than do we need any of them at all. 

Not sure if you remember anything more about this?

PR #1020, opened 2024-01-15
Add sass build time monitoring [OR-483]

notlee:
Let's build a case to invest in our Sass and its tooling. Or, better, to [proactively move away from sass](https://docs.google.com/document/d/1RuGduWdX0zGsgsp9C7lIhXgqEia6sWK900_3XVwYDIM/edit#heading=h.1f3yolavobef)!

Completed todo:

1. Do the Platforms team think this is a good idea? 💚 
2. Option to turn off 🟢 
3. [Helpful link](https://github.com/Financial-Times/origami/pull/1407) on why this is here, and what we'll do about it 🟢 – will release when this is approved
4. Capture [number of sass compilation errors](https://github.com/webpack-contrib/sass-loader/blob/03773152760434a2dd845008c504a09c0eb3fd91/src/index.js#L68) 🟠  – this is a big enough PR, we can add this later if we want
5. Remote, aggregate monitoring through [biz-ops metrics api](https://github.com/Financial-Times/biz-ops-metrics-api)  🟢  - [example](https://biz-ops.in.ft.com/api-explorer?loggedInUser=kiya.gurmesa&query=%7B%0A++teams%28where%3A+%7B+code%3A+%22platforms-customer-products%22+%7D%29+%7B%0A++++delivers%28where%3A+%7B+code%3A+%22page-kit%22+%7D%29+%7B%0A++++++code%0A++++++metric%28name%3A+%22sass-build-time%22%29+%7B%0A++++++++value%0A++++++++timestamp%0A++++++%7D%0A++++%7D%0A++++metricRollup%28path%3A+%22delivers.metric%22%2C+method%3A+sum%29+%7B%0A++++++value%0A++++++timestamp%0A++++%7D%0A++%7D%0A%7D&title=Custom+Biz+Ops+report)

### DIFF
--- a/packages/dotcom-build-sass/src/monitored-sass-loader.ts
+++ b/packages/dotcom-build-sass/src/monitored-sass-loader.ts
@@ -4,8 +4,19 @@ import os from 'os'
 import sassLoader from 'sass-loader'
 
 class SassStats {
-  #noticeThrottle = 10 * 1000
-  #stats = { totalTime: 0, notice: null }
+  #noticeStrategies = ['throttle', 'never', 'always']
+  #noticeStrategy = this.#noticeStrategies.includes(process.env.FT_SASS_STATS_NOTICE)
+    ? process.env.FT_SASS_STATS_NOTICE
+    : 'throttle'
+  #noticeThrottleSeconds =
+    typeof process.env.FT_SASS_STATS_NOTICE_THROTTLE_SECONDS === 'number'
+      ? process.env.FT_SASS_STATS_NOTICE_THROTTLE_SECONDS
+      : 60 * 60 * 0.5 // show throttled notice given 30 mins since last notice
+  #noticeThrottlePercentage =
+    typeof process.env.FT_SASS_STATS_NOTICE_THROTTLE_PERCENTAGE === 'number'
+      ? process.env.FT_SASS_STATS_NOTICE_THROTTLE_PERCENTAGE
+      : 30 // show throttled notice given a 30% increase
+  #stats = { totalTime: 0, noticeDate: null, totalTimeAtLastNotice: 0 }
   #directory = path.join(os.tmpdir(), 'dotcom-build-sass')
   #file = path.join(this.#directory, 'sass-stats.json')
   #startTime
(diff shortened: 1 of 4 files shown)


notlee:
Rhys pointed me to a webpack plugin which can monitor times for loaders in a more general way. It can output as json and it's possible to provide a callback function. This might be helpful if we want to measure total build time. This would allow us to ensure as sass build time decreases, the total build does too. 

Looks like there are a few issues open with compatibility complaints though, so it might be one to come back to. I'm aware of introducing any issues, so generic build time monitoring might be something to come back to. Right now I'm interested in drawing attention to the cost of Sass. https://github.com/stephencookdev/speed-measure-webpack-plugin

joelcarr:
> Do the Platforms team think this is a good idea?

Platforms team think this is a good idea 🙌

notlee:
I've added environment variables so we can change the behaviour of reporting independently of PageKit releases.

akomiqaia (commented):
LGTM

packages/dotcom-build-sass/src/monitored-sass-loader.ts
```diff
@@ -4,8 +4,19 @@ import os from 'os'
 import sassLoader from 'sass-loader'
 
 class SassStats {
-  #noticeThrottle = 10 * 1000
-  #stats = { totalTime: 0, notice: null }
+  #noticeStrategies = ['throttle', 'never', 'always']
+  #noticeStrategy = this.#noticeStrategies.includes(process.env.FT_SASS_STATS_NOTICE)
```
joelcarr, line 17: **question:** How will these environment variables be set?

notlee: I intend to work with product teams to set via Doppler for their projects. Run this idea past Umberto who is happy to do that for Storytelling :)

joelcarr: Lovely! Makes sense to me.

joelcarr (commented):
Please could we add your team to the `CODEOWNERS` for the path `packages/dotcom-build-sass/`?

joelcarr (approved):
Wonderful! Excited (and dreading) to see the data caught with this!

PR #1022, opened 2024-01-18
CPP-1757: loosen appContext.contentType schema

apaleslimghost:
loosens the schema for `appContext.contentType` to accept any string. this allows us to use both the current (CAPI-style, PascalCase) values and the old (ES-style, kebab-case) values, and prevents implicitly coupling this schema to the list of possible content types that CAPI can return.

i don't like that the schema validator can throw an error if the app context is invalid, when some of it is provided by data we don't control. but i'm not about to rewrite how this works 😅

### DIFF
--- a/packages/dotcom-server-app-context/src/schema.json
+++ b/packages/dotcom-server-app-context/src/schema.json
@@ -6,49 +6,71 @@
     "abTestState": {
       "type": "string",
       "description": "The A/B test flags data as a comma delimited string",
-      "examples": ["subscriberCohort:on,premiumCohort:on,nonUSACohort:on"],
+      "examples": [
+        "subscriberCohort:on,premiumCohort:on,nonUSACohort:on"
+      ],
       "pattern": "^,*([0-9A-Za-z-_]+:[0-9A-Za-z-_]+,*)+$"
     },
     "appName": {
       "type": "string",
       "description": "The name of the application",
-      "examples": ["front-page", "stream-page", "article-page"],
+      "examples": [
+        "front-page",
+        "stream-page",
+        "article-page"
+      ],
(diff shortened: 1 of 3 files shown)


packages/dotcom-server-app-context/src/schema.json
```diff
@@ -6,49 +6,74 @@
...
       "description": "The type or sub-type of the content on the current page",
-      "examples": ["article", "video", "audio", "podcast", "package", "live-blog"],
-      "pattern": "^(article|video|audio|podcast|package|live-blog)$"
+      "examples": [
+        "article",
+        "video",
+        "audio",
+        "podcast",
+        "package",
+        "live-blog"
```
joelcarr, line 66: **todo:** We should update the examples

apaleslimghost: oops, missed it

packages/dotcom-server-app-context/src/schema.json
```diff
@@ -6,49 +6,74 @@
...
+      ],
+      "pattern": "^(Article|Placeholder|Video|Audio|LiveBlogPackage|LiveBlogPost|ContentPackage|Content|MediaResource)$"
```
apaleslimghost, line 68: **thought**: it occurs to me that this implicitly ties Page Kit to cp-content-pipeline, and if we ever add another content type, we'd need to update this, and it won't be obvious that we need to. maybe we should make the schema more lenient.

joelcarr (approved):

cerdfied (approved):
lgtm other than the examples joel mentioned!

cerdfied (approved):

PR #1026, opened 2024-02-27
fix:  don't load polyfill.io in shell

apaleslimghost:
[#inc-2575-polyfill-io-is-intermittently-failing](https://financialtimes.enterprise.slack.com/archives/C06LT31R3BM)

[#polyfill-discussion-2024](https://financialtimes.enterprise.slack.com/archives/C06LNUG3C12)

although this should probably be a major change, we'll release it as a minor to make the upgrade path easier. also tbh nothing really needs polyfills any more?

there is some cleanup work to do after this to remove the `dotcom-ui-polyfill-service` package (going with the quickest fix for now)

### DIFF
--- a/packages/dotcom-ui-shell/src/components/Shell.tsx
+++ b/packages/dotcom-ui-shell/src/components/Shell.tsx
@@ -12,7 +12,6 @@ import {
 } from '@financial-times/dotcom-ui-base-styles'
 import { FlagsEmbed, TFlagsEmbedProps } from '@financial-times/dotcom-ui-flags'
 import { Bootstrap, TBootstrapProps } from '@financial-times/dotcom-ui-bootstrap'
-import * as polyfillService from '@financial-times/dotcom-ui-polyfill-service'
 import formatAttributeNames, { TAttributeData } from '../lib/formatAttributeNames'
 import GTMHead from './GTMHead'
 import GTMBody from './GTMBody'
@@ -37,7 +36,6 @@ function Shell(props: TShellProps) {
   }
 
   const resourceHints = [
-    polyfillService.enhanced(),
     // There is no need to include stylesheets here as any <link rel="stylesheet" /> tags
     // should be found by the browser's speculative parser.
     ...props.scripts,
--- a/packages/dotcom-ui-shell/src/components/Shell.tsx
+++ b/packages/dotcom-ui-shell/src/components/Shell.tsx
@@ -32,8 +32,8 @@ type TShellProps = TDocumentHeadProps &
 
 function Shell(props: TShellProps) {
   const bootstrapProps: TBootstrapProps = {
-    coreScripts: [polyfillService.core()],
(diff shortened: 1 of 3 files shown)


packages/dotcom-ui-shell/src/components/Shell.tsx
```diff
@@ -32,8 +32,8 @@ type TShellProps = TDocumentHeadProps &
 
 function Shell(props: TShellProps) {
   const bootstrapProps: TBootstrapProps = {
-    coreScripts: [polyfillService.core()],
-    enhancedScripts: [polyfillService.enhanced(), ...props.scripts]
+    coreScripts: [],
+    enhancedScripts: props.scripts
   }
 
   const resourceHints = [
```
ivomurrell, line 39: **question**: it's okay that that there's still a resource hint for polyfill.io, right?

ivomurrell (approved):

packages/dotcom-ui-shell/src/components/Shell.tsx
```diff
@@ -32,8 +32,8 @@ type TShellProps = TDocumentHeadProps &
 
 function Shell(props: TShellProps) {
   const bootstrapProps: TBootstrapProps = {
-    coreScripts: [polyfillService.core()],
-    enhancedScripts: [polyfillService.enhanced(), ...props.scripts]
+    coreScripts: [],
```
GlynnPhillips, line 35: **Question:** You planning to do the clean up of `@financial-times/dotcom-ui-polyfill-service` afterwards?

packages/dotcom-ui-shell/src/components/Shell.tsx
apaleslimghost: let's remove it just in case

packages/dotcom-ui-shell/src/components/Shell.tsx
GlynnPhillips: Does [line 15](https://github.com/Financial-Times/dotcom-page-kit/pull/1026/files#diff-07a5a1f5a61d8361b19ebcc04b34f55c76561580f67451783776e7d7e4523d8cR15) need removing to avoid unused warnings?

apaleslimghost: yea

GlynnPhillips: Does[ line 40](https://github.com/Financial-Times/dotcom-page-kit/pull/1026/files#diff-07a5a1f5a61d8361b19ebcc04b34f55c76561580f67451783776e7d7e4523d8cR40) need removing?

packages/dotcom-ui-shell/src/components/Shell.tsx
ivomurrell: should be fiiiine seeing as preloading doesn't execute any code but I suppose it would be good to stop fetching it seeing as we don't need it

ivomurrell (approved):

PR #1031, opened 2024-04-18
Add ASK FT button

Ventzy:
## What
- Add `ASK FT` button linking to https://ask.ft.com in `dotcom-ui-header` at:
  - Top header - visible on screen size `L` and up
  - Drawer - visible on smaller screens, same as search in the drawer
  - Sticky header - visible on screen size `L` and up
- Add new optional `showAskButton` prop in `THeaderOptions` to control `ASK FT` button rendering. By default, the button is not rendered which keeps FT header unchanged for current applications using it. `ASK FT` button will be selectively made visible per application by explicitly setting `showAskButton` to `true`.
- Add `showAskButton` prop in the stories
- Update README
- Add Top, Drawer and Sticky snapshot tests with `showAskButton: true`

## Ticket
https://financialtimes.atlassian.net/browse/ARP-86

## Design
- Desktop/Top header - https://www.figma.com/file/G0vA3biIFmIeXa0bB3bqWW/Advanced-Research-Tools?type=design&node-id=3934-60616&mode=design&t=Rb1z6BXxlJJM9ekX-4
- Mobile/Drawer - https://www.figma.com/file/G0vA3biIFmIeXa0bB3bqWW/Advanced-Research-Tools?type=design&node-id=3986-60964&mode=design&t=MYIjwsXoshgQTUF9-4

### Top header, storybook with `showAskButton: true`
image

### Drawer, storybook with `showAskButton: true`
image

### Drawer, storybook with `showAskButton: true`, mobile resolution
image

### Sticky header, storybook with `showAskButton: true`
image


apaleslimghost (dismissed):
hi @Ventzy, thanks for this! a couple of things:

- has this feature been discussed with product leadership in Customer Products? adding something to the site-wide header is a very impactful change and i want to make sure the right people are informed
- for experimental features we typically prefer them not to be built in Page Kit directly, to ensure it's stable and reduce churn. a common pattern for this kind of thing is for Page Kit to provide a component slot in the appropriate templates and apps to pass the new component in. sorry this wasn't brought up when you were directed to make the change in Page Kit, i'll make sure the team are aware of this and update the contribution guidelines to make it clear 🙂

packages/dotcom-ui-header/src/header.scss
```diff
@@ -62,4 +62,46 @@
 		$size: 32
 	);
   vertical-align: middle;
-}
\ No newline at end of file
+}
+
+.o-header__drawer-search-term {
+  min-width: 120px;
```
joelcarr, line 68: **question:** Is this required? When looking at this in Storybook, I don't see it having an effect. It is best practice not to override Origami classes so trying to avoid it if we can.

Ventzy: Unfortunately, the input width is not shrunk bellow certain width on resolutions between 981 and 1319 without it. I've tested on latest Chrome and Firefox on Windows and I've just double checked it again

image

This is closest explanation I've found https://stackoverflow.com/questions/42421361/input-button-elements-not-shrinking-in-a-flex-container

`min-width` solves the issue and I've choose 120px because this is a bit less than a minimum width I see this input can go. When I've played with it I've tried different approaches trying to avoid setting this - I've tried different html structure, css grid instead of flexbox, but they all involved much more html and css changes and the input didn't want to shrink anyway.

I've just tested an alternative solution with setting `width: 100%` instead of `min-width: 120px` as suggested in another SO comment. It is working and we will avoid the magic number. I will change it if you agree with that approach.

joelcarr: Thank you for the thorough explanation! I agree that we should swap to the `width: 100%` rule instead 👍

Ventzy: Updated. Here are few tests on my end

### With ASK FT button
#### Desktop
https://github.com/Financial-Times/dotcom-page-kit/assets/1334199/cada6ce6-0646-4bd2-81ae-4e53938a8a6d

#### Mobile
image


### Without ASK FT button (keep unchanged)
#### Desktop
https://github.com/Financial-Times/dotcom-page-kit/assets/1334199/e50e72dd-92b4-4fa1-849e-2f2432e9f5a7

#### Mobile
image

joelcarr (approved):

joelcarr:
Well GitHub didn't like me submitting a reason for dismissing @apaleslimghost's review with a proper sentence 🤷‍♂️ Here is what I wrote originally:

> We have discussed this approach with the team and for this instance we approve the implementation.

PR #1042, opened 2024-05-22
Replace sticky header myFT link with My Account / Sign in

GlynnPhillips:
### Description

Replace the myFT logo with the My Account or Sign in links in the sticky header

### Justification

This change reflects similar changes made to the main header. We now want My Account to be more of a hub that can take you to places like myFT so promoting this from this point in the navigation makes more sense.

### Screen recording

https://github.com/Financial-Times/dotcom-page-kit/assets/524573/fc53f0d0-d706-4fff-9c05-e7b16c83ec8d

### Ticket

https://financialtimes.atlassian.net/browse/ACC-3083

### DIFF
--- a/packages/dotcom-ui-header/src/__test__/components/__snapshots__/StickyHeader.spec.tsx.snap
+++ b/packages/dotcom-ui-header/src/__test__/components/__snapshots__/StickyHeader.spec.tsx.snap
@@ -235,15 +235,14 @@ exports[`dotcom-ui-header/src/components/StickyHeader renders ASK FT button 1`] 
             Subscribe
           </a>
           <a
-            className="o-header__top-icon-link o-header__top-icon-link--myft "
-            data-trackable="my-ft"
-            href="/myft"
+            className="o-header__top-myaccount"
+            data-trackable="My Account"
+            href="/myaccount"
+            id="o-header-top-link-myaccount"
             tabIndex={-1}
           >
-            <span
-              className="o-header__visually-hidden"
-            >
-              myFT
+            <span>
+              My Account
             </span>
           </a>
         </div>
@@ -540,15 +539,14 @@ exports[`dotcom-ui-header/src/components/StickyHeader renders as a logged in use
             Subscribe
           </a>
           <a
-            className="o-header__top-icon-link o-header__top-icon-link--myft "
-            data-trackable="my-ft"
-            href="/myft"
+            className="o-header__top-myaccount"
+            data-trackable="My Account"
+            href="/myaccount"
+            id="o-header-top-link-myaccount"
             tabIndex={-1}
           >
-            <span
-              className="o-header__visually-hidden"
-            >
-              myFT
+            <span>
+              My Account
             </span>
           </a>
         </div>
@@ -834,7 +832,27 @@ exports[`dotcom-ui-header/src/components/StickyHeader renders as an anonymous us
         </div>
         <div
           className="o-header__top-column o-header__top-column--right"
-        />
+        >
+          <a
+            className="o-header__top-button o-header__top-button--hide-m"
+            data-trackable="Subscribe"
+            href="/products?segmentId=#"
+            role="button"
+            tabIndex={-1}
+          >
+            Subscribe
+          </a>
+          <a
+            className="o-header__top-myaccount"
+            data-trackable="Sign In"
+            href="/login?location=#"
+            id="o-header-top-link-signin"
+          >
+            <span>
+              Sign In
+            </span>
+          </a>
+        </div>
       </div>
     </div>
   </div>
--- a/packages/dotcom-ui-header/src/__test__/output/component.spec.tsx
+++ b/packages/dotcom-ui-header/src/__test__/output/component.spec.tsx
@@ -95,12 +95,12 @@ describe('dotcom-ui-header', () => {
 
       expect(
         container.querySelector(
-          '.o-header__top-column .o-header__top-column--right a[data-trackable="Subscribe"]'
+          '.o-header__top-column.o-header__top-column--right a[data-trackable="Subscribe"]'
         )
       ).not.toBeNull()
       expect(
         container.querySelector(
-          '.o-header__top-column .o-header__top-column--right a[data-trackable="Sign In"]'
+          '.o-header__top-column.o-header__top-column--right a[data-trackable="Sign In"]'
         )
       ).not.toBeNull()
     })
--- a/packages/dotcom-ui-header/src/components/sticky/partials.tsx
+++ b/packages/dotcom-ui-header/src/components/sticky/partials.tsx
@@ -2,7 +2,7 @@
 /* This is the sticky header variant */
 
 import React from 'react'
-import { SubscribeButton, SignInLink } from '../top/partials'
+import { TopColumnRight } from '../top/partials'
 import { THeaderProps } from '../../interfaces'
 import { AskFtButton } from '../ask-ft/askFtButton'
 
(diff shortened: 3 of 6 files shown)


packages/dotcom-ui-header/src/__test__/output/component.spec.tsx
```diff
@@ -95,12 +95,12 @@ describe('dotcom-ui-header', () => {
...
       expect(
         container.querySelector(
-          '.o-header__top-column .o-header__top-column--right a[data-trackable="Subscribe"]'
+          '.o-header__top-column.o-header__top-column--right a[data-trackable="Subscribe"]'
         )
       ).not.toBeNull()
       expect(
         container.querySelector(
-          '.o-header__top-column .o-header__top-column--right a[data-trackable="Sign In"]'
+          '.o-header__top-column.o-header__top-column--right a[data-trackable="Sign In"]'
```
GlynnPhillips, line 103: **Reviewer Note:** I am intrigued by this test, my changes seem to cause it to fail but I can't understand why or how it ever worked 🤔 

`.o-header__top-column .o-header__top-column--right` should have never worked because as far as I can tell they have always been the same element and never a parent/child relationship - which also makes sense because the second class is a modifier so it shouldn't be a child.

Here are places from main prior to any of my changes that show it being the same element.

https://github.com/Financial-Times/dotcom-page-kit/blob/3dadb7b1986ece05d75f38d5a4598cbbd0445bd1/packages/dotcom-ui-header/src/components/top/partials.tsx#L139

https://github.com/Financial-Times/dotcom-page-kit/blob/3dadb7b1986ece05d75f38d5a4598cbbd0445bd1/packages/dotcom-ui-header/src/components/top/partials.tsx#L215

https://github.com/Financial-Times/dotcom-page-kit/blob/3dadb7b1986ece05d75f38d5a4598cbbd0445bd1/packages/dotcom-ui-header/src/components/sticky/partials.tsx#L81


https://github.com/Financial-Times/dotcom-page-kit/blob/3dadb7b1986ece05d75f38d5a4598cbbd0445bd1/packages/dotcom-ui-header/src/components/sticky/partials.tsx#L155

packages/dotcom-ui-header/src/components/sticky/partials.tsx
```diff
@@ -132,31 +106,8 @@ const TopColumnCenterSticky = (props: THeaderProps) => {
...
-  let children: JSX.Element | undefined = undefined
-
-  if (props.userIsLoggedIn) {
-    children = <NavListRightLoggedInSticky {...props} />
-  } else if (props.showUserNavigation) {
-    children = <NavListRightAnonSticky {...props} />
-  }
-
-  return <div className="o-header__top-column o-header__top-column--right">{children}</div>
+  return <TopColumnRight {...props} />
```
GlynnPhillips, line 110: **Reviewer note:** Theoretically this `TopColumnRightSticky` component is no longer needed as the place it is used from could just use `TopColumnRight` directly. However in the place that `TopColumnRightSticky` is used ([here](https://github.com/Financial-Times/dotcom-page-kit/pull/1042/files#diff-66f378e331d22f88bd01c75c962869f3fe338ebcb2ae2a1f2482762522439989R71-R79)) it is building up a component out of `sticky` components, so I felt it kept a better consistency to keep this component as a proxy to `TopColumnRight`

packages/dotcom-ui-header/src/interfaces.d.ts
```diff
@@ -17,4 +17,4 @@ export type THeaderProps = THeaderOptions & {
   data: TNavigationData
 }
 
-export type THeaderVariant = 'simple' | 'large-logo'
+export type THeaderVariant = 'simple' | 'large-logo' | 'sticky'
```
GlynnPhillips, line 20: **Reviewer note:** The `sticky` variant allows us to easily reuse the same top right components for the my account / sign in and subscribe links in the sticky header as the main header but whilst also varying the `tabIndex` attribute in order to remove the link from the tab order when navigating by keyboards

packages/dotcom-ui-header/src/__test__/components/__snapshots__/StickyHeader.spec.tsx.snap
```diff
@@ -834,7 +832,27 @@ exports[`dotcom-ui-header/src/components/StickyHeader renders as an anonymous us
...
+            className="o-header__top-myaccount"
+            data-trackable="Sign In"
+            href="/login?location=#"
+            id="o-header-top-link-signin"
+          >
+            <span>
+              Sign In
+            </span>
+          </a>
+        </div>
```
GlynnPhillips, line 855: **Review notes:** This appears to have been missing from the original snapshot output. Based on my investigation I am not sure why 🤷‍♂️ Looking at the code and also the experience of the sticky header on production I believe its correct that these should be shown in this sceneario (Anon user)

packages/dotcom-ui-header/src/__test__/output/component.spec.tsx
joelcarr: Figured it out! The `NavListRightAnonSticky` [component contains a div with the classes](https://github.com/Financial-Times/dotcom-page-kit/blob/c8ec521523337c4b5cbb69c88a79b2099a0bda14/packages/dotcom-ui-header/src/components/sticky/partials.tsx#L82) `o-header__top-column o-header__top-column--right`. This component is rendered within the `TopColumnRightSticky` [component that has a div with the same classes](https://github.com/Financial-Times/dotcom-page-kit/blob/c8ec521523337c4b5cbb69c88a79b2099a0bda14/packages/dotcom-ui-header/src/components/sticky/partials.tsx#L156-L159). That is why these tests were passing.

joelcarr: We can see this HTML being rendered in this screenshot:

image

joelcarr: As your new changes reflects the rendered HTML in the Origami component, I am confident in the changes made here and this test now better reflecting the expected markup.

packages/dotcom-ui-header/src/components/sticky/partials.tsx
joelcarr: I feel like I have come across this before. I agree it is best to keep this as a proxy for now.

packages/dotcom-ui-header/src/__test__/output/component.spec.tsx
GlynnPhillips: Ohh great find. Thanks Joel! Looks like someone added accidental nesting due to the split of the code into granular functions/components

packages/dotcom-ui-header/src/__test__/components/__snapshots__/StickyHeader.spec.tsx.snap
GlynnPhillips: Ok, I now have an answer for this (I missed something in my original investigation that a nights sleep helped with)

The fixture data that is passed into the test sets `userIsLoggedIn` to `false` and `showUserNavigation` to `false` which means that [this if statement](https://github.com/Financial-Times/dotcom-page-kit/blob/main/packages/dotcom-ui-header/src/components/sticky/partials.tsx#L153-L157) never matches and therefore renders nothing within the top right nav.

On ft.com I don't believe that this is a scenario we actually use, when a user is anonymous we still show right column nav to prompt the user to sign in or subscribe. So I feel like my changes are now a better representation of what we use in production

PR #1043, opened 2024-05-22
fix: Update o-header peer dependency

GlynnPhillips:
In v9.5.0 I managed to miss updating the peer dependency for o-header.

This ensure thats the correct version is now being used in v10

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -59,8 +59,8 @@
         "typescript": "4.1.6"
       },
       "engines": {
-        "node": "16.x || 18.x || 20.x",
-        "npm": "7.x || 8.x || 9.x || 10.x"
+        "node": "18.x || 20.x",
+        "npm": "8.x || 9.x || 10.x"
       }
     },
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -59,8 +59,8 @@
         "typescript": "4.1.6"
       },
       "engines": {
-        "node": "16.x || 18.x || 20.x",
-        "npm": "7.x || 8.x || 9.x || 10.x"
+        "node": "18.x || 20.x",
+        "npm": "8.x || 9.x || 10.x"
```
GlynnPhillips, line 63: **Note:** @joelcarr I am guessing in v10 we didn't update all the package-locks - do you want me to include this in this change or would prefer me to take it out and only include the o-header change?

joelcarr: 🤦‍♂️ I did forget to do that! Yes please 🙏

joelcarr (approved):

PR #1045, opened 2024-05-23
Replace sticky header myFT link with My Account / Sign in - v9

GlynnPhillips:
📢 This is a copy of https://github.com/Financial-Times/dotcom-page-kit/pull/1042 but for v9.x

### Description

Replace the myFT logo with the My Account or Sign in links in the sticky header

### Justification

This change reflects similar changes made to the main header. We now want My Account to be more of a hub that can take you to places like myFT so promoting this from this point in the navigation makes more sense.

### Screen recording

https://github.com/Financial-Times/dotcom-page-kit/assets/524573/fc53f0d0-d706-4fff-9c05-e7b16c83ec8d

### Ticket

https://financialtimes.atlassian.net/browse/ACC-3083


joelcarr (approved):
:fishsticks:

PR #1048, opened 2024-07-09
CON-3426: update o-header dependency

danieleruiz:
# Description

Update o-header dependency to decrease the width of the search bar

# Ticket
https://financialtimes.atlassian.net/browse/CON-3426

# ScreenShot
image

### DIFF
--- a/packages/dotcom-ui-header/package.json
+++ b/packages/dotcom-ui-header/package.json
@@ -29,15 +29,14 @@
     "@financial-times/logo-images": "^1.10.1",
     "@svgr/core": "^5.0.0",
     "camelcase": "^6.0.0",
-    "check-engine": "^1.10.1",
-    "@financial-times/o-header": "^13.0.0"
+    "check-engine": "^1.10.1"
   },
   "peerDependencies": {
-    "@financial-times/o-header": "^13.0.0",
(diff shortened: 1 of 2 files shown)


packages/dotcom-ui-header/package.json
```diff
@@ -29,15 +29,14 @@
...
     "@svgr/core": "^5.0.0",
     "camelcase": "^6.0.0",
-    "check-engine": "^1.10.1",
-    "@financial-times/o-header": "^13.0.0"
+    "check-engine": "^1.10.1"
   },
   "peerDependencies": {
-    "@financial-times/o-header": "^13.0.0",
     "@financial-times/logo-images": "^1.10.1",
+    "@financial-times/o-header": "13.0.1",
```
metart43, line 36: o-header is a peer dependency now. Is this intended?

metart43: Oh, it was removed from `devDependencies` and it's referenced in two places 🤔

metart43 (approved):

packages/dotcom-ui-header/package.json
danieleruiz: it only needs to be peer dependency

joelcarr (approved):

PR #1049, opened 2024-07-10
CON-3426: fix padding on ask to ft in header

danieleruiz:
# Description
in the new search bar design on header we had the issue with ask to ft button that the style looks incorrect
this PR fixed it
it fix too the problems with typeahead results box

# ScreenShot
Before:
Captura de pantalla 2024-07-10 a las 14 54 18
image
image

After:
image
image
image

### DIFF
--- a/packages/dotcom-ui-header/src/header.scss
+++ b/packages/dotcom-ui-header/src/header.scss
@@ -34,6 +34,10 @@
 .o-header__drawer-search-term {
   width: 100%;
 }
+.o-header__top-wrapper .ft-header__ask-ft-button {
+  padding: 8px;
+  margin-left: -12px;
+}
 .ft-header__ask-ft-button {
   @include oTypographySans(-1, $weight: 'semibold');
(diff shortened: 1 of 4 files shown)


packages/dotcom-ui-header/src/header.scss
```diff
@@ -34,7 +34,10 @@
 .o-header__drawer-search-term {
   width: 100%;
 }
-
+.o-header__top-wrapper .ft-header__ask-ft-button {
+  padding: 8px;
+  margin-left: -12px;
```
joelcarr, line 39: **question:** Has this been tested in all devices sizes? I am not a fan of negative values for layout rules unless it is absolutely necessary?

danieleruiz: the new structure moves a little bit the ask ft button to the right but maybe is something that we can assume and avoid negative margin

metart43:
I QAed this and your fixes work fine. Thanks

joelcarr (approved):

PR #1053, opened 2024-08-27
chore: update myft wording in tests

GlynnPhillips:
# Description

Update the link text for myFT in the tests

# Justification

The link text was recently changed in the origami navigation service as these tests use real data these also need updating.

Original navigation service change https://github.com/Financial-Times/origami-navigation-service/pull/976


metart43 (approved):

GlynnPhillips:
**Question/Thought:** I have made the simple change for now, but I wonder how much these tests should be integration tests that are testing the values that come from external services? It seems like these services might be better off mocked so changing the origami navigation service doesn't mean you need to know that tests in this repo will fail as a result. However I am not sure how tricky that would be in this project and if it has very little reward given that the navigation service changes in frequently (I assume)?

joelcarr (approved):

PR #1052, opened 2024-08-27
[CON-3348] o-header and typeahead

metart43:
# Note 

This PR includes changes from https://github.com/Financial-Times/dotcom-page-kit/pull/1051. It has been rebased and merge conflicts has been resolved. These two features has to be released at the same time. Please, see the very first commit https://github.com/Financial-Times/dotcom-page-kit/pull/1052/commits/7f8be8ea084c447db46d503396938ae11521c017

cc: @Ventzy 

# Description

1. Mark-up changes to accommodate `o-header` v.14.0.1. (https://github.com/Financial-Times/origami/pull/1777)
1.1 Changing input search field to `type="search"`. This should improve the screen reader experience, also reflect true sematic meaning of the input field.
1.2 Input search field now has [`role="combobox"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/combobox_role) to indicate that an element as an input that controls another element, such as a listbox or grid, that can dynamically pop up to help the user set the value of that input. 
In our case - typeahead (https://github.com/Financial-Times/n-topic-search/releases/tag/v7.0.0)
1.3 `aria-controls` attribute references typeahead wrapper
1.4 Close and Search buttons have undergone slight changes.
2. The initialization of the Typeahead component has been updated. Previously, it was implemented as a plain JavaScript class, it was re-written as a Preact functional component.
3. Added a bit of testing around the new mark-up.
4. New typeahead is written in Preact, so it seems like it's getting bundled as well. That's why the ci was failing previously. https://github.com/Financial-Times/dotcom-page-kit/pull/1052/commits/ede633b6034654eb4d971cc28a5cc1e8b20fbdca
5. Safari version was bumped accordingly to our [browser support policy](https://docs.google.com/document/d/1z6kecy_o9qHYIznTmqQ-IJqre72jhfd0nVa4JMsS7Q4/edit#heading=h.er7no6jwfim1)
6. `o-header` [fix](https://github.com/Financial-Times/origami/pull/1797/files) contained optional chaining syntax. That's why I needed to make a change to the webpack config and make sure webpack understands it.

# Demos 

<details>
<summary>
Interactivity demo [mobile]
</summary>

https://financialtimes.enterprise.slack.com/files/U04T4CBAV6C/F07GK51SZSP/screenrecording_08-12-2024_16-39-20_1.mp4

</details>

<details>
<summary>
Interactivity demo [desktop]
</summary>

[...]

PR #1056, opened 2024-09-04
Add preact compat to avoid errors importing dotcom-ui-header

danieleruiz:
# Description
We have detected an "React undefined" error occurring when updating the `dotcom-ui-layout > dotcom-ui-header` component in certain projects, such as **next-barrier-page**. This issue is due to the new typeahead feature, which uses Preact and was introduced in this pull request:
[Financial-Times/dotcom-page-kit#1052](https://github.com/Financial-Times/dotcom-page-kit/pull/1052)
This PR explicitly adds preact/compat to resolve it

Slack conversation:
https://financialtimes.slack.com/archives/C010BC61NDS/p1725461070723719


apaleslimghost:
hi @danieleruiz, i don't think this is quite the right way to fix this, give me a bit of time to work out what the correct fix is 🙂

danieleruiz:
> hi @danieleruiz, i don't think this is quite the right way to fix this, give me a bit of time to work out what the correct fix is 🙂

thanks @apaleslimghost , maybe another option is to add an alias to preact in tsconfig?

apaleslimghost:
that won't work, the `paths` option is only for Typescript's module resolution, it doesn't affect the code that's output.

danieleruiz:
> that won't work, the `paths` option is only for Typescript's module resolution, it doesn't affect the code that's output.
adding this two properties too
https://preactjs.com/guide/v10/typescript/
 "jsx": "react-jsx",
 "jsxImportSource": "preact",
but i think that going to make work preact components but going to make fail react components.
i dont know if dotcom-ui-header can have their own tsconfig

apaleslimghost:
the `tsconfig` options aren't going to work on their own, as the `browser.js` isn't actually compiled by Typescript. i think getting that working would be to much effort, so for now, the easiest solution is going to be using `h` directly in this file, instead of JSX, e.g.:

```diff
-render(<TypeAhead container={mobileDrawer} inputId={input.id} />, typeaheadContainer)
+render(h(TypeAhead, {container: mobileDrawer, inputId: input.id}), typeaheadContainer)
```

we should also add `preact` as an explicit dependency of `dotcom-ui-header`. 

in the future i would like us to not be using Preact as direct dependencies of component packages, as choosing between React or Preact should be done at an application level. i'm writing a proposal for that.

danieleruiz:
> the `tsconfig` options aren't going to work on their own, as the `browser.js` isn't actually compiled by Typescript. i think getting that working would be to much effort, so for now, the easiest solution is going to be using `h` directly in this file, instead of JSX, e.g.:
> 
> ```diff
> -render(<TypeAhead container={mobileDrawer} inputId={input.id} />, typeaheadContainer)
> +render(h(TypeAhead, {container: mobileDrawer, inputId: input.id}), typeaheadContainer)
> ```
> 
> we should also add `preact` as an explicit dependency of `dotcom-ui-header`.
> 
> in the future i would like us to not be using Preact as direct dependencies of component packages, as choosing between React or Preact should be done at an application level. i'm writing a proposal for that.

hello @apaleslimghost i tested it using npm link and it didnt work for me.. do you have a working PR with this?

danieleruiz:
> the `tsconfig` options aren't going to work on their own, as the `browser.js` isn't actually compiled by Typescript. i think getting that working would be to much effort, so for now, the easiest solution is going to be using `h` directly in this file, instead of JSX, e.g.:
> 
> ```diff
> -render(<TypeAhead container={mobileDrawer} inputId={input.id} />, typeaheadContainer)
> +render(h(TypeAhead, {container: mobileDrawer, inputId: input.id}), typeaheadContainer)
> ```
> 
> we should also add `preact` as an explicit dependency of `dotcom-ui-header`.
> 
> in the future i would like us to not be using Preact as direct dependencies of component packages, as choosing between React or Preact should be done at an application level. i'm writing a proposal for that.

hello @apaleslimghost i tested it using npm link and it didnt work for me.. do you have a working PR with this?

kavanagh:
Thanks for taking a look at this @apaleslimghost! Just one thought on this...

> in the future i would like us to not be using Preact as direct dependencies of component packages, as choosing between React or Preact should be done at an application level. i'm writing a proposal for that.

We could just go all in on Preact on the client? It's decent, used on app and dotcom and already a dependency anyway on a number of the big app, notable on article page iirc... so maybe just yolo the preact think and happy days 😀 ?!

packages/dotcom-ui-header/package.json
```diff
@@ -23,7 +23,8 @@
   "license": "MIT",
   "dependencies": {
     "@financial-times/dotcom-types-navigation": "file:../dotcom-types-navigation",
-    "n-topic-search": "^8.0.0"
+    "n-topic-search": "^8.0.0",
+    "preact": "^10.23.2"
```
joelcarr, line 27: **issue:** We should add this as a peer dependency so that a consuming system will use a version of preact that matches the version range (if it already has a version.)

danieleruiz: changed!

danieleruiz: finally the class was not need and the close text should be shown

joelcarr (approved):

PR #1057, opened 2024-09-05
Revert "[CON-3348] o-header and typeahead"

joelcarr:
Reverts Financial-Times/dotcom-page-kit#1052


apaleslimghost:
i discussed this with @alexmuller, we don't think there's any need to revert this

PR #1062, opened 2024-09-16
merge n-ui-foundations into dotcom-ui-base-styles

apaleslimghost:
this PR moves all the sass from [n-ui-foundations](https://github.com/Financial-Times/n-ui-foundations/) into `dotcom-ui-base-styles`. since this includes all the git history from n-ui-foundations, there's a lot of commits, and it's not going to be easily reviewable on its own. i'm going to do the actual work of integrating this sass into `dotcom-ui-base-styles` in a separate PR based on this one, to make that easier to review.


ivomurrell (approved):
**nitpick**: this would be better as a merge commit rather than rebased on top of the current history

PR #1063, opened 2024-09-16
integrate ex-n-ui-foundations styles into dotcom-ui-base-styles

apaleslimghost:
updates the n-ui-foundations styles included in #1062 to fit more into the Page Kit ecosystem, with an O3 future in mind. notably, this removes support for using `dotcom-ui-base-styles` in components, and recommends that components use Origami styles directly (there is no longer a need to have a default bundled set of styles for both components and apps, this is handled via `peerDependencies`).

this PR starts consuming `o3-foundation` as a dependency, to replace `nUiZIndexFor` with `o3-foundation`'s z-index variables. there's no way to import _just_ the z-index variables, so this imports the whole of `o3-foundation`. until we migrate `dotcom-ui-base-styles` fully to o3, this will increase the CSS bundle size. that's a tradeoff i'm willing to make for the next few months, to facilitate an easier migration path.


packages/dotcom-ui-base-styles/browser.js
```diff
@@ -1,11 +1,5 @@
 import oTypography from '@financial-times/o-typography'
 
-// Polyfill for :focus-visible https://github.com/WICG/focus-visible
-// NOTE: v5 of this polyfill is not yet supported by o-normalise
-// https://github.com/WICG/focus-visible/pull/196/files
-// https://github.com/Financial-Times/o-normalise/issues/41
-import 'focus-visible'
```
ivomurrell, line 7: **todo**: we can uninstall the `focus-visible` package now

ivomurrell (approved):
**question**:
> this removes support for using dotcom-ui-base-styles in components

should we remove the `component.js` entrypoint then?

apaleslimghost:
> should we remove the component.js entrypoint then?

it's misleadingly named, that entrypoint isn't for use in components, it's the entrypoint that _exports_ a [JSX component](https://github.com/Financial-Times/dotcom-page-kit/blob/base-styles-rethink/packages/dotcom-ui-base-styles/src/components/LoadFontsEmbed.tsx) for server rendering

PR #1090, opened 2024-10-23
Switch source of GTM based on flag

oliverturner:
## Description

This PR enables [First Party mode](https://developers.google.com/tag-platform/tag-manager/first-party/setup-guide?setup=manual) for Google Tag Manager (GTM): when the `ads-first-party-gtm` Toggler flag is enabled the GTM script is loaded from `ft.com/page-resources` rather than `www.googletagmanager.com/gtm.js`

This should offer us some protection against anti-tracking measures that block requests to `googletagmanager.com`

## Reviewing

Reviewing commit-by-commit is recommended. The PR comprises 2:
1. The substantive changes
2. Snapshot updates


packages/dotcom-ui-shell/src/components/GTMBody.tsx
```diff
@@ -7,10 +7,14 @@ const GTMBody = ({ flags }: { flags: TFlagsData }) => {
     return null
   }
 
+  const src = flags['ads-first-party-gtm']
+    ? 'https://www.ft.com/page-resources/'
```
apaleslimghost, line 11: **question**: is there a reason `GTMHead` uses a relative URL and this uses absolute?

oliverturner: None whatsoever... thanks for catching!

PR #1091, opened 2024-10-28
ADSDEV 2064: GTM first party mode

oliverturner:
> [!Note]
> This PR supersedes the deleted PR `ADSDEV-2064/proxy-gtm-requests`, reducing its scope to updating the source of the  script tag in the HEAD

## Description

This PR enables [First Party mode](https://developers.google.com/tag-platform/tag-manager/first-party/setup-guide?setup=manual) for Google Tag Manager (GTM): when the `ads-first-party-gtm` Toggler flag is enabled the GTM script is loaded from `ft.com/page-resources` rather than `www.googletagmanager.com/gtm.js`

This should offer us some protection against anti-tracking measures that block requests to `googletagmanager.com`

## Reviewing

Reviewing commit-by-commit is recommended. The PR comprises 2:
1. The substantive changes
2. Snapshot updates

### DIFF
--- a/packages/dotcom-ui-shell/src/components/GTMHead.tsx
+++ b/packages/dotcom-ui-shell/src/components/GTMHead.tsx
@@ -7,11 +7,17 @@ const GTMHead = ({ flags }: { flags: TFlagsData }) => {
     return null
   }
 
-  const tagManager = `(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
-          new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
-            j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
-            'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
-        })(window,document,'script','dataLayer','GTM-NWQJW68');`
+  const src = flags['ads-first-party-gtm']
+    ? 'https://www.ft.com/page-resources'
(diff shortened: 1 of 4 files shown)


packages/dotcom-ui-shell/src/components/GTMHead.tsx
```diff
@@ -7,11 +7,17 @@ const GTMHead = ({ flags }: { flags: TFlagsData }) => {
     return null
   }
 
-  const tagManager = `(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
-          new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
-            j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
-            'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
-        })(window,document,'script','dataLayer','GTM-NWQJW68');`
+  const src = flags['ads-first-party-gtm']
+    ? '/page-resources'
```
joelcarr, line 11: **issue:** This should be an absolute URL as currently this path will only work in production. It will cause issues for consuming apps in their non production environments.

joelcarr (changes requested):

packages/dotcom-ui-shell/src/components/GTMHead.tsx
oliverturner: That's an excellent point 👍

joelcarr (approved):

PR #1123, opened 2025-01-22
CPP-2222 Use Webpack's native import resolution for S/CSS and drop Postcss

ivomurrell:
# Description

With Webpack 5 we've been able to update to a newer version of the `sass-loader` loader. The newer version [no longer](https://github.com/webpack-contrib/sass-loader/releases/tag/v11.0.0) requires imports to be prefixed with a `~` if they want to use Webpack's resolution logic (i.e., traversing `node_modules`.) This is particularly good for stylesheets that import stylesheets that themselves import stylesheets, as the transitive imports are no longer required to use the non-standard `~` prefix to work.

This means we no longer need to use the `postcss-import` plugin to work around the import behaviour. Importing using `sass-loader` makes native use of Webpack's resolution logic so should be better equipped to handle de-duping etc., though that has not been thoroughly tested: the main motivation here is to use fewer plugins and be more native.

Now that we no longer use `postcss-import`, we no longer require PostCSS at all. Let's remove it entirely from the Sass plugin (though it's still used transitively by the `css-loader`.)

# Checklist

- [x] This feature is stable, i.e. is not an ongoing experiment, temporary workaround, or hack

### DIFF
--- a/packages/dotcom-build-sass/src/index.ts
+++ b/packages/dotcom-build-sass/src/index.ts
@@ -48,13 +48,6 @@ export class PageKitSassPlugin {
     }
 
     const postcssLoaderOptions = {
-      postcssOptions: {
-        plugins: [
-          // Allow @import of CSS files from node_modules
-          // https://github.com/postcss/postcss-import
-          require('postcss-import')()
-        ]
-      },
       implementation: require('postcss')
     }
 
--- a/packages/dotcom-build-sass/src/index.ts
+++ b/packages/dotcom-build-sass/src/index.ts
@@ -42,8 +42,8 @@ export class PageKitSassPlugin {
       sassOptions: {
         // Disable formatting so that we don't spend time pretty printing
         outputStyle: 'compressed',
-        // Enable Sass to @import source files from installed dependencies
-        includePaths: ['node_modules/@financial-times', 'node_modules', ...this.includePaths]
+        // Enable Sass to @import source files from additional relative paths
+        includePaths: this.includePaths
       }
     }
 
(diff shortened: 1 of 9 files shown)


packages/dotcom-build-sass/src/index.ts
```diff
@@ -43,25 +43,14 @@ export class PageKitSassPlugin {
         // Disable formatting so that we don't spend time pretty printing
         outputStyle: 'compressed',
         // Enable Sass to @import source files from installed dependencies
-        includePaths: ['node_modules/@financial-times', 'node_modules', ...this.includePaths]
+        includePaths: [...this.includePaths]
```
apaleslimghost, line 46: **suggestion**: there's no need to clone `includePaths`, so the spread is unnecessary

```suggestion
        includePaths: this.includePaths
```

packages/dotcom-build-sass/src/index.ts
```diff
@@ -43,25 +43,14 @@ export class PageKitSassPlugin {
         // Disable formatting so that we don't spend time pretty printing
         outputStyle: 'compressed',
         // Enable Sass to @import source files from installed dependencies
```
apaleslimghost, line 45: **nitpick**: this comment isn't accurate any more as we don't have a default `includePaths`

packages/dotcom-build-sass/src/index.ts
```diff
@@ -91,12 +80,6 @@ export class PageKitSassPlugin {
           loader: require.resolve('css-loader'),
           options: cssLoaderOptions
         },
-        // Enable use of PostCSS for CSS postprocessing
-        // https://github.com/postcss/postcss-loader
-        {
-          loader: require.resolve('postcss-loader'),
-          options: postcssLoaderOptions
-        },
```
apaleslimghost, line 99: **praise**: you love to see it

apaleslimghost (approved):
noice

PR #1129, opened 2025-02-05
Include ex-n-ui-foundations Sass files in dotcom-ui-base-styles' npm package

ivomurrell:
# Description

The new `sass` directory was not added to the `files` filter in the `package.json` so wasn't being included in npm releases.

# Checklist

- [x] This feature is stable, i.e. is not an ongoing experiment, temporary workaround, or hack

### DIFF
--- a/packages/dotcom-ui-base-styles/package.json
+++ b/packages/dotcom-ui-base-styles/package.json
@@ -34,6 +34,7 @@
   "files": [
     "dist/",
     "src/",
+    "sass/",
     "browser.js",
     "component.js",
     "styles.scss"


apaleslimghost (approved):
**suggestion**: i think we should switch from `files` to `.npmignore`, would be more intuitive, and the denylist approach is more common in other packages

ivomurrell:
> **suggestion**: i think we should switch from `files` to `.npmignore`, would be more intuitive, and the denylist approach is more common in other packages

agreed, i would have spent a lot longer trying to work out why this wasn't working if you hadn't pointed me in the right direction. i looked for `.npmignore` and was confused when i couldn't find any

PR #1136, opened 2025-02-13
fix: don't revert to core experience on error from third-party script

apaleslimghost:
https://financialtimes.enterprise.slack.com/archives/C08CDEJCVCN

### DIFF
--- a/packages/dotcom-ui-bootstrap/lib/bootstrap.js
+++ b/packages/dotcom-ui-bootstrap/lib/bootstrap.js
@@ -25,9 +25,15 @@
       console.error('The script ' + script + ' failed to load') // eslint-disable-line no-console
     }
 
+    const scriptHost = new URL(script).hostname
+
     if (/enhanced/.test(doc.className)) {
-      console.warn('Script loading failed, reverting to core experience') // eslint-disable-line no-console
-      doc.className = doc.className.replace('enhanced', 'core')
+      if (scriptHost === 'www.ft.com') {
+        console.warn('Script loading failed, reverting to core experience') // eslint-disable-line no-console
+        doc.className = doc.className.replace('enhanced', 'core')
+      } else {
+        console.warn('Third-party script, not reverting to core experience') // eslint-disable-line no-console
+      }
     }
 
     if (scriptsConfig.trackErrors) {
--- a/packages/dotcom-ui-bootstrap/README.md
+++ b/packages/dotcom-ui-bootstrap/README.md
@@ -125,7 +125,7 @@ If JavaScript is available the `no-js` class on the document element will be rep
 
 If the browser passes the [cuts the mustard](#cutting-the-mustard) test then the `core` class name on the document element will be replaced with `enhanced`.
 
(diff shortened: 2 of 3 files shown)


oliverturner (commented):
Smart hotfix! 

Approving in advance of snapshot test update 👍

oliverturner (approved):

packages/dotcom-ui-bootstrap/lib/bootstrap.js
```diff
@@ -25,9 +25,15 @@
       console.error('The script ' + script + ' failed to load') // eslint-disable-line no-console
     }
 
+    const scriptHost = new URL(script).hostname
+
     if (/enhanced/.test(doc.className)) {
-      console.warn('Script loading failed, reverting to core experience') // eslint-disable-line no-console
-      doc.className = doc.className.replace('enhanced', 'core')
+      if (scriptHost === 'www.ft.com') {
```
olusoladavid, line 31: **question:** Do we need to update the [documentation here](https://github.com/Financial-Times/dotcom-page-kit/blob/main/packages/dotcom-ui-bootstrap/README.md) to reflect this breaking change to the behavior of **required** enhanced scripts?

Also, what do you think about introducing a `thirdPartyScripts` prop to the Pagekit Shell to make a feature distinction between required enhanced scripts and non-required third-party scripts?

oliverturner: Making use of `coreScripts` would be my approach here:
https://github.com/Financial-Times/dotcom-page-kit/blob/5bbdc87306bb2e6970a86b739cb0e1c08ad1ff3f/packages/dotcom-ui-shell/src/components/Shell.tsx#L33-L36

apaleslimghost: that sounds like a good idea long term, but it's going to require a lot of changes in apps to start making use of `thirdPartyScripts`, so i think we should just go for this hotfix for now

olusoladavid: Happy to approve and revisit. Thank you

apaleslimghost: > Do we need to update the [documentation here](https://github.com/Financial-Times/dotcom-page-kit/blob/main/packages/dotcom-ui-bootstrap/README.md) to reflect this breaking change

@olusoladavid this isn't a breaking change, it's slightly more permissive behaviour than before. i'll put a note in the readme though

olusoladavid (approved):
Thanks for the quick fix ⭐

packages/dotcom-ui-bootstrap/lib/bootstrap.js
olusoladavid: Thanks @apaleslimghost . I was thinking from the perspective of an API consumer who might have already designed for this behavior where all script load failures, regardless of origin, would cause a reversion to core experience. That thing they say about people using every observable behavior of your API as part of your public API

PR #1154, opened 2025-04-01
Add links fetching to Dropdown component for Pro Navigation

pdimova:
# Description
Jira: [ENT-2257](https://financialtimes.atlassian.net/browse/ENT-2257)

This PR aims to enhance the functionality of the Dropdown component for users who have JavaScript enabled.

The enhancement is carried out in two steps:
	1.	Fetch the links from the pro-navigation service. The returned JSON follows the same format as the default rendered links.
	2.	Recreate the list of link items using the newly fetched data.

If the user does not have JavaScript enabled, or if there is an issue or error when fetching the data, they will see the default list rendered instead.

How to debug: The entry point is the newly created function updateProNavigationLinks()

If the user does not have JavaScript enabled, or if there is an issue or error when fetching the data, they will see the default list rendered instead.

Screenshot 2025-04-02 at 11 06 51

The Vanilla Js approach:

There are multiple ways to render the fetched data. This PR uses direct DOM manipulation with vanilla JavaScript and built-in browser APIs.
This approach is performance-conscious and avoids the need to import any additional packages.

Other possible approaches include:

1.	Using React’s createRoot/render functionalities
This approach results in cleaner, more maintainable code.
The drawback is the need to check the installed React version as a peer dependency and dynamically import the appropriate render functions. From my observations, some repositories consuming dotcom-ui-header are still using React 17.

2.	Creating and reusing a string template
This involves defining a string template that can be reused both within the Dropdown component and on the client side when iterating over the newly fetched data.
It is easier to maintain since the template is defined in a single place.
However, this would require the Dropdown component to accept a string template via the dangerouslySetInnerHTML prop, which is not ideal.

3.	Using an external package such as react-element-to-jsx-string
This would allow us to convert the Dropdown list component into a reusable template, keeping the required HTML in a single location, making future updates simpler.
The downside is that it introduces an additional dependency to the header component and may require compatibility with newer versions of React, as some of these packages rely on React 19.

4.	Converting the Dropdown component to Preact
This is also a viable option, particularly since Preact is already imported in the browser.js file and is used to render the TypeAhead component.
However, this would require changes to the TypeScript and bundler configuration.

We welcome further discussion on any of the approaches outlined above.

**Metrics**
Performance metrics were collected for four key areas in the code:
	•	Fetching the links
	•	Equality check between default and fetched links
	•	DOM manipulations for updating the list via the dropdown (we have two instances – one in the header and another in the sticky header)
	•	Overall performance of the `updateProNavigationLinks` function:

**Code setup:**
Screenshot 2025-04-08 at 10 20 04

_Performance test tool_: Chrome DevTools v134.0.6998.166 (Official Build) (arm64)
_Test machine:_ Mac , Apple M1, 16GB ram, Sequoia OS 15.4
_Network throttling_: Slow 4G
_CPU throttling_: 4x slowdown
_Fetching_ from local.ft.com/api/links (Express server running locally)

**Results:**
Screenshot 2025-04-08 at 10 30 11

Screenshot 2025-04-08 at 10 16 56

Screenshot 2025-04-08 at 10 18 50

Part 1: https://github.com/Financial-Times/dotcom-page-kit/pull/1150
# Checklist

[ENT-2257]: https://financialtimes.atlassian.net/browse/ENT-2257?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -1,93 +1,211 @@
...
-        closeDropdownButton.style.pointerEvents = 'auto'
-        // When the close button is clicked, reset the dropdown button so we can open the dropdown
-        // correctly next time we click the dropdown button.
-        // Also blur the close dropdown button (the mobile close button) to close the dropdown
-        closeDropdownButton.addEventListener('click', () => {
-            dropdownButton.removeAttribute('data-dropdown-button-active')
-            closeDropdownButton.blur()
-        })
+      const dropdownButton = dropdownContainer.querySelector('.o-header__dropdown-button')
+      if (!dropdownContainer.contains(event.target)) {
```
alex-ruskov-ft, line 59: this would also fire if you click anywhere on the dropdown not only the buttons, which we want to specifically track

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -1,93 +1,211 @@
...
+
+const initProDropdown = async () => {
+  const proDropdowns = document.querySelectorAll('.o-header__dropdown.pro_navigation');
+  if (!proDropdowns) {
+    console.error('Navigation dropdown not found');
+    return;
+  }
+
+  try {
+    const links = await fetchLinks('https://pro-navigation.ft.com/api/links');
```
alex-ruskov-ft, line 102: I am not sure if page-kit has fetches anywhere else, but may be better to keep the url in an env variable?

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -1,93 +1,211 @@
...
+    if (!validateLinks(links) || isEqual(PRO_NAVIGATION_DROPDOWN_DEFAULT_LIST, links)) {
+      return;
+    }
+
+    proDropdowns.forEach((dropdown) => updateLinksList(dropdown, links));
+  }
+  catch (error) {
+    console.error('Error updating dropdown navigation:', error);
```
alex-ruskov-ft, line 110: I think we shouldn't log the original error in the console, maybe there is some other logger which could be utilised instead of logging in the browser?

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -1,93 +1,211 @@
...
+  const response = await fetch(url);
+  if (!response.ok) {
+    throw new Error(`HTTP error! Status: ${response.status}`);
+  }
+  return response.json();
+};
+
+const validateLinks = (links) => {
+  if (!Array.isArray(links) || links.length === 0) {
+    console.error('Invalid links data format or empty links list');
```
alex-ruskov-ft, line 124: theoretically this could throw an error rather than just log, as anyways the call is in a try catch.

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -1,93 +1,211 @@
...
+    return false;
+  }
+  return true;
+};
+
+const updateLinksList = (dropdown, links) => {
+  const list = dropdown.querySelector('.o-header__dropdown-list');
+  const listItem = dropdown.querySelector('.o-header__dropdown-list-item');
+  if (!list) {
```
alex-ruskov-ft, line 133: **nitpick**: we can do this check before querying for the list item, just to save the DOM traversal. Also could we use the first child of the list for the listItem rather than looking for a selector?

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -1,93 +1,211 @@
...
+  }
+
+  const label = new DocumentFragment();
+  const siblingNodes = Array.from(listItem.querySelector('a').children).slice(1); // Skip the first child (list-item-details-container)
+  siblingNodes.forEach(node => label.append(node));
+
+  return label;
+}
+
+const buildListItem = (template, link) => {
```
alex-ruskov-ft, line 163: **Suggestion:** I think the listItem and the label should be separate parameters, since we don't use TS here and it may be easy to break stuff.

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -1,93 +1,211 @@
...
+  if (!listItem) {
+    console.error('List item template not found');
+    return;
+  }
+
+  const listItemClone = listItem.cloneNode(true);
+  const a = listItemClone.querySelector('a');
+  const icon = a.querySelector('.o-header__dropdown-icon');
+  const span = a.querySelector('span:not([class])');
+  const labelClone = label.cloneNode(true);
```
alex-ruskov-ft, line 174: Maybe we can have a check if the label exists before we attempt to call cloneNode on it?

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -1,93 +1,211 @@
...
+
+  if (!listItemClone || !a || !icon || !span) {
+    console.error('Invalid template structure');
+    return;
+  }
+
+  // Update class name
+  listItemClone.className = `o-header__dropdown-list-item ${link.hasBottomLine ? 'o-header__dropdown-list-divider' : ''}`;
```
alex-ruskov-ft, line 182: **Question:** Is there a way we could either filter or push to the classname list instead of directly setting the className, considering that in this way we wouldn't need to mention any other classnames apart from the divider?
Also should it be .className and not .class since we are already in vanilla html/JS?

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -1,93 +1,211 @@
...
+
+  // Update link attributes
+  a.setAttribute('href', link.href);
+
+  const match = a.getAttribute('data-trackable')?.match(/^(.*)_.*_clicked$/);
+  const trackingKey = match ? match[1] : 'pro-navigation';
```
alex-ruskov-ft, line 188:
```suggestion
  const trackingKey = match ? match[1] : 'pro_navigation';
```

alex-ruskov-ft: Also match[1] may not work if we have something like pro_navigation_clicked, as I think match[1] would just be pro

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -1,93 +1,211 @@
...
+  a.setAttribute('data-trackable', `${trackingKey}_${link.id}_clicked`);
+
+  // Update icon and text
+  icon.className = `o-header__dropdown-icon ${link.hasAccess ? link.icon + '-icon' : 'lock-icon'}`;
```
alex-ruskov-ft, line 192:
```suggestion
  icon.className = `o-header__dropdown-icon ${link.hasAccess ? link.icon : 'lock'}-icon`;
```

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -1,93 +1,211 @@
...
+  if (link.hasLabel) {
+    a.appendChild(labelClone);
+  }
+
+  return listItemClone;
+};
+
+const init = () => {
+  enhanceInteractivity()
+
```
alex-ruskov-ft, line 205: **Suggestion:** Maybe just add a small comment that any other dropdown navigation use-case specific initialisations are to be listed below.

packages/dotcom-ui-header/src/__test__/components/dropdown-navigation/utils.spec.js
alex-ruskov-ft, line 1: We could leave only one of the two files (this or the one below)

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -1,93 +1,211 @@
...
+  const trackingKey = match ? match[1] : 'pro-navigation';
+  a.setAttribute('data-trackable', `${trackingKey}_${link.id}_clicked`);
+
+  // Update icon and text
+  icon.className = `o-header__dropdown-icon ${link.hasAccess ? link.icon + '-icon' : 'lock-icon'}`;
+  span.innerText = link.title;
+
+  // Append label
+  if (link.hasLabel) {
+    a.appendChild(labelClone);
```
phelipemaia, line 197: **suggestion:** instead of making DOM changes, wouldn't be better to create a fragment and update all at once? https://developer.mozilla.org/en-US/docs/Web/API/Document/createDocumentFragment

pdimova: Maybe you mean this: https://github.com/Financial-Times/dotcom-page-kit/pull/1154/files#diff-a3d6c74ceedae642b7abccd37ae8e0792423e8d7c6a85d742867cdc35c18287fR140

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
pdimova: Good idea! I will leave it out of scope for this iteration of the review.

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
pdimova: I agree – unfortunately, I don’t see any logger being used across the other packages, only console logging the error itself. I’ll just keep the message for now and check with the team to see what their preferred approach is for handling errors.

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -1,93 +1,211 @@
...
+}
+
+const extractLabel = (listItem) => {
+  if (!listItem) {
+    console.error('List item template not found');
+    return;
+  }
+
+  const label = new DocumentFragment();
+  const siblingNodes = Array.from(listItem.querySelector('a').children).slice(1); // Skip the first child (list-item-details-container)
```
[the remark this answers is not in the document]
pdimova: Good catch! I have updated the logic for handling the label.

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
pdimova: className is the fastest way. But we can toggle it as welll

packages/dotcom-ui-header/src/__test__/components/__snapshots__/MainHeader.spec.tsx.snap
```diff
@@ -2103,7 +2103,7 @@ exports[`dotcom-ui-header/src/components/MainHeader renders FT Pro Dropdown menu
                   </a>
                 </li>
                 <li
-                  className="o-header__dropdown-list-item false"
+                  className="o-header__dropdown-list-item "
```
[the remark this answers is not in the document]
pdimova: It is due to ternary operator check setting empty string when false. I have updated the logic for this though

packages/dotcom-ui-header/src/__test__/components/dropdown-navigation/DropdownNavigation.spec.tsx
```diff
@@ -4,13 +4,14 @@
 import React from 'react'
 import { render } from '@testing-library/react'
 import '@testing-library/jest-dom/extend-expect'
-import { DropdownNavigation } from '../../components/dropdown-navigation/dropdownNavigation'
-import {PRO_NAVIGATION_DROPDOWN_DEFAULT_LIST} from '../../components/dropdown-navigation/dropdownNavigationDefaultLists'
+import { DropdownNavigation } from '../../../components/dropdown-navigation/dropdownNavigation'
+import {PRO_NAVIGATION_DROPDOWN_DEFAULT_LIST} from '../../../components/dropdown-navigation/constants'
```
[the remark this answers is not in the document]
pdimova: Will let Prettier to handle the spacing and indentation.

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -1,93 +1,196 @@
...
-                }
-            }
-        },
-        { threshold: [0.01] }
-    )
-
-    stickyHeaderObserver.observe(stickyHeader)
+  // Determine the list item template
+  const label = dropdown.querySelector('.o-header__dropdown-list-item-label-container')
+  const listItem = label?.closest('.o-header__dropdown-list-item') || list.firstElementChild
```
alex-ruskov-ft, line 130: I like the use of closest here so we don't traverse the whole tree but only the ancestors

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -1,93 +1,196 @@
...
+    return
+  }
+
+  // Update class name
+  listItemClone.classList.toggle('o-header__dropdown-list-divider', link.hasBottomLine)
+
+  // Update link attributes
+  a.setAttribute('href', link.href)
+
+  const match = a.getAttribute('data-trackable')?.match(/^(.*)_.*_clicked$/)
```
alex-ruskov-ft, line 171: What would happen if the link id which is a part of the key contains 1 or more "_"? Wouldn't this break the tracking key extraction?

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -1,93 +1,196 @@
...
+  const trackingKey = match ? match[1] : 'pro_navigation'
+  a.setAttribute('data-trackable', `${trackingKey}_${link.id}_clicked`)
+
+  // Update icon and text
+  icon.className = `o-header__dropdown-icon ${link.hasAccess ? link.icon : 'lock'}-icon`
```
alex-ruskov-ft, line 176: **Question:** Is there a way we can only specify the icon class and not need to specify o-header__dropdown-icon? Maybe using toggle? In this way even if o-header__dropdown-icon is changed/removed from the tsx, we would be immune to it

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -1,93 +1,196 @@
...
+  span.innerText = link.title
+
+  if (label && !link.hasLabel) {
```
alex-ruskov-ft, line 179: **Question**: What would happen if the current list doesn't have an item with a label, but the fetched list asks for one?

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
pdimova: For broader cases like the ones below, regex alone would not be enough. We either go for specific delimiters if possible or another approach like data-attribute would be better
`tracking_key_other_string_here_link_id_begins_clicked`
`antoher_example_key_many_other_strings_and_now_the_link_here_clicked`

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
pdimova: Toggle might not help, you can check the updated logic for avoiding extra class names

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
pdimova: Good question!

This scenario might be possible, and it’s best to handle it alongside the question raised about the use of the hasLabel key. https://github.com/Financial-Times/pro-navigation/pull/8#discussion_r2029254985
It's worth it a separate ticket and further discussion.

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
alex-ruskov-ft: I am okay with the data attribute idea which would keep the clean key

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -1,93 +1,208 @@
...
+    }
+
+    proDropdowns.forEach((dropdown) => updateLinksList(dropdown, links, trackingKey))
+  } catch (error) {
+    console.error('Error updating dropdown navigation.')
+  }
+}
+
+const fetchLinks = async (url) => {
+  const response = await fetch(url)
```
olusoladavid, line 116: **curious:** why does the endpoint respond with a 5xx error code when the session token is missing. I would expect a 401 or similar? Also, nit: could the message just be `Missing session token` and less specific about cookie name?
image

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -1,93 +1,208 @@
...
-                        document.activeElement.blur()
-                    }
-                }
-            }
-        },
-        { threshold: [0.01] }
-    )
-
-    stickyHeaderObserver.observe(stickyHeader)
+const updateLinksList = (dropdown, links, trackingKey) => {
```
olusoladavid, line 123: **light-suggestion:** Given the lack of typing, adding the `El` suffix makes it easy to quickly figure which args are HTML elements
```suggestion
const updateLinksList = (dropdownEl, links, trackingKey) => {
```

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -1,93 +1,208 @@
...
+  const listItem = label?.closest('.o-header__dropdown-list-item') || list.firstElementChild
+
+  if (!listItem) {
+    console.error('List item template not found')
+    return
+  }
+
+  // Build the updated list
+  const updatedListFragment = document.createDocumentFragment()
+  links.forEach((link) => {
```
olusoladavid, line 141: **question:** I understand the argument for wanting to be able to dynamically modify the link list on the backend. But there are two aspects I am curious about:
- If the links API is down for some reason, does it mean that the users will not be able to discover and navigate to protected pages at all?
- how do you guard against site-wide UI regression given that dropdown may behave differently at certain list lengths?

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -1,93 +1,208 @@
...
+    const updatedItem = buildListItem(listItem, label, link, trackingKey)
+    updatedItem && updatedListFragment.append(updatedItem)
+  })
+
+  // Update the DOM
+  list.innerHTML = ''
+  list.append(updatedListFragment)
```
olusoladavid, line 148: possible convenience method?
```suggestion
  list.replaceChildren(updatedListFragment)
```

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -1,93 +1,208 @@
...
+const init = () => {
+  enhanceInteractivity()
+
+  // Add any dropdown navigation-specific initializations below.
+  // Make sure to use a unique selector to distinguish your dropdown from others.
+  updateProNavigationLinks({
+    selector: 'pro_navigation',
+    trackingKey: 'pro_navigation',
+    defaultLinks: PRO_NAVIGATION_DROPDOWN_DEFAULT_LIST,
+    proNavigationApi: 'https://pro-navigation.ft.com/api/links'
```
olusoladavid, line 202: **question:** Is this API endpoint versioned? How do we plan to handle breaking changes to the API considering that different apps will usually be using different/older versions of PageKit?

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.tsx
```diff
@@ -1,8 +1,10 @@
 import React from 'react'
 
 export interface DropdownNavigationProps {
+  selector: string
+  buttonId?: string
```
olusoladavid, line 5: **suggestion:** would be nice to have some form of documentation that explains each of these props

olusoladavid (commented):
Great work @pdimova ! Had a first run through the implementation and it was quite easy to follow. I had some questions/comments. Will test against a live app tomorrow and add any final review. Thank you!

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
pdimova: `replaceChildren` is safe to use in modern browsers from 2020 onwards. Can we afford this? 
We went with backward compatibility for browsers from around 2017-2019,  which is why we avoided relatively modern CSS features like :has().

olusoladavid: Thanks for the context @pdimova . It does make sense to try to support these older browsers where possible. However, based on our [browser support matrix/policy,](https://docs.google.com/document/d/1z6kecy_o9qHYIznTmqQ-IJqre72jhfd0nVa4JMsS7Q4/edit?usp=sharing) I think 2020+ features fit perfectly. I don't have any strong preference either way. `replaceChildren` just looked more convenient and might have a more performant implementation than first clearing the node and then doing an `append`.

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -1,93 +1,208 @@
...
+
+  try {
+    const links = await fetchLinks(proNavigationApi)
+    if (!Array.isArray(links) || links.length === 0 || isEqual(defaultLinks, links)) {
+      return
+    }
+
+    proDropdowns.forEach((dropdown) => updateLinksList(dropdown, links, trackingKey))
+  } catch (error) {
+    console.error('Error updating dropdown navigation.')
```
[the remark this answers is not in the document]
pdimova: Although the querySelector function would fail gracefully if no element is found, and we do check for various missing elements, I’d prefer to retain the catch for unexpected DOM operation errors. This is due to the nature of the implementation: a reusable, somewhat generic Dropdown component, with separate client-side logic that is not part of the React component itself.

I have added the suggested tracking events.

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -1,93 +1,208 @@
...
+
+  const proDropdowns = document.querySelectorAll(`.o-header__dropdown.${selector}`)
+  if (!proDropdowns) {
+    console.error('Navigation dropdown not found')
+    return
+  }
+
+  try {
+    const links = await fetchLinks(proNavigationApi)
+    if (!Array.isArray(links) || links.length === 0 || isEqual(defaultLinks, links)) {
```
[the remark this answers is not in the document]
pdimova: For small (to medium-sized) objects, using JSON.stringify should be perfectly acceptable.

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
pdimova: > If the links API is down for some reason, does it mean that the users will not be able to discover and navigate to protected pages at all (even when they have access)?

Valid responses from the pro-navigation API will be cached in Fastly for up to 30 minutes. If the API does not respond after this period, your statement will be valid indeed.
By default, all link objects will be rendered as if the user has access to them (apart from the Subscription Manager URL which is not shown initially). If the user does not actually have access, they will simply be directed to a landing page, as I understand it.

> how do you guard against site-wide UI regression given that dropdown may behave differently at certain list lengths?

The dropdown has both a minimum and maximum height. If the list of links exceeds the maximum height, the dropdown will become scrollable

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
pdimova: As this is still at an experimental stage, we would like to see the outcome before proceeding with API versioning.

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.tsx
pdimova: JsDocs description added

pdimova: hey @olusoladavid , do you know where I could add some documentation/prop description for the Dropdown component itself

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
olusoladavid: guess we'll do more testing once we are ready to rollout, thank you!

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
olusoladavid: thank you. I imagine there's a ticket for this followup work already?

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.tsx
olusoladavid: I think adding a README within the Dropdown component directory might be a good place to start. Thanks for the JSDoc! Already very helpful

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
pdimova: @olusoladavid It’ll definitely be part of the upcoming planning, though, so we’ll make sure it’s captured then.

olusoladavid (approved):
LGTM! Thanks for the great work and absolute diligence

phelipemaia (approved):

PR #1161, opened 2025-04-24
Add component view event to the Dropdown component

pdimova:
# Description

This PR adds support for dispatching component view events for the Dropdown content.

Context
As part of an ongoing experiment, the tracking functionality is implemented directly within the Dropdown component to avoid introducing changes for consumers who do not have view tracking enabled.

# Checklist
	•	Component view event dispatched from the JS init function.
	•	Remove tracking attributes from Dropdown jsx

Notes
This is an experimental addition and may be refactored or removed based on the outcomes of the tracking initiative.

- [ ] I've discussed this feature with [the Platforms team](https://financialtimes.enterprise.slack.com/archives/C3TJ6KXEU)
- [ ] My branch has been rebased onto the latest commit on `main` (don't merge `main` into your branch)

### DIFF
--- a/packages/dotcom-ui-header/src/__test__/components/__snapshots__/StickyHeader.spec.tsx.snap
+++ b/packages/dotcom-ui-header/src/__test__/components/__snapshots__/StickyHeader.spec.tsx.snap
@@ -580,7 +580,6 @@ exports[`dotcom-ui-header/src/components/StickyHeader renders FT Pro dropdown me
             <div
               className="o-header__dropdown-content"
               data-id="dropdown-content"
-              data-o-tracking-view="pro_navigation_component_view"
               id="dropdown-options"
               role="group"
               tabIndex={-1}
--- a/packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
+++ b/packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
@@ -97,6 +97,44 @@ const enhanceInteractivity = () => {
   stickyHeaderObserver.observe(stickyHeader)
 }
 
+const trackContent = (options) => {
+  if (!window.IntersectionObserver) {
+    return
+  }
+
+  const selector = options && options.selector
+  const elementsToTrack = document.querySelectorAll(selector)
+  if (elementsToTrack.length === 0) {
+    return
+  }
+
+  const onChange = (changes, observer) => {
+    changes.forEach((change) => {
+      if (change.isIntersecting || change.intersectionRatio > 0) {
+        const viewedEl = change.target
+        const eventData = {
+          action: 'view',
+          category: 'component',
+          className: viewedEl.className,
+          componentContentId: viewedEl.dataset.id,
+          url: window.document.location.href || null,
+          nodeName: viewedEl.nodeName,
+          href: viewedEl.href || false,
+          text: viewedEl.text || false,
+          role: viewedEl.role || false
+        }
+
+        document.body.dispatchEvent(new CustomEvent('oTracking.event', { detail: eventData, bubbles: true }))
+        observer.unobserve(viewedEl)
+      }
+    })
+  }
+
+  const observer = new IntersectionObserver(onChange, { threshold: [1.0] })
+
+  elementsToTrack.forEach((el) => observer.observe(el))
+}
+
 const updateProNavigationLinks = async (options) => {
   const { selector, trackingKey, defaultLinks, proNavigationApi } = options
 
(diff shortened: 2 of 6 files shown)


packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -207,8 +242,17 @@ const buildListItem = (listItem, label, link, trackingKey) => {
 const init = () => {
   enhanceInteractivity()
 
+  /**
+   * Dispatches a custom event with tracking data when the dropdown content becomes visible.
```
alex-ruskov-ft, line 246: Should we put this comment above the function definition instead of the call?

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -97,6 +97,44 @@ const enhanceInteractivity = () => {
   stickyHeaderObserver.observe(stickyHeader)
 }
 
+const trackContent = (options) => {
+  if (!window.IntersectionObserver) {
+    return
+  }
+
+  const selector = options && options.selector
```
alex-ruskov-ft, line 105: **Question:** Do we need an options object, can't we just have one parameter called selector?

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -97,6 +97,44 @@ const enhanceInteractivity = () => {
   stickyHeaderObserver.observe(stickyHeader)
 }
 
+const trackContent = (options) => {
```
alex-ruskov-ft, line 100: **Suggestion:** Maybe we can change the name of the function to hint it is component view tracking. Something like trackDropdownView?

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -97,6 +97,44 @@ const enhanceInteractivity = () => {
...
+          role: viewedEl.role || false
+        }
+
+        document.body.dispatchEvent(new CustomEvent('oTracking.event', { detail: eventData, bubbles: true }))
+        observer.unobserve(viewedEl)
+      }
+    })
+  }
+
+  const observer = new IntersectionObserver(onChange, { threshold: [1.0] })
```
alex-ruskov-ft, line 133: Not sure if threshold of [1.0] is a good idea, since the event may never fire if you are on a smaller screen where the dropdown exceeds the screen height. What do you think?

packages/dotcom-ui-header/src/__test__/components/__snapshots__/StickyHeader.spec.tsx.snap
```diff
@@ -580,7 +580,6 @@ exports[`dotcom-ui-header/src/components/StickyHeader renders FT Pro dropdown me
             <div
               className="o-header__dropdown-content"
               data-id="dropdown-content"
-              data-o-tracking-view="pro_navigation_component_view"
```
alex-ruskov-ft, line 583: Do we have an equivalent of this value in the event data that we emit?

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
pdimova: I simply prefer passing a single object over multiple arguments—also, consistency.

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
pdimova: The default threshold is 1.0, which works in most cases. We can lower it to handle edge cases with small screens, lots of top content, and slightly larger lists.

packages/dotcom-ui-header/src/__test__/components/__snapshots__/StickyHeader.spec.tsx.snap
pdimova: According to the documentation, this attribute is only used to make the element trackable and does not contribute any values to the event data. The additional data is being collected by the getContextData function.

[Ref](https://github.com/Financial-Times/origami/tree/main/libraries/o-tracking#otrackingviewinit)

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
alex-ruskov-ft: Fair enough , I am just always cautious when we use js and not typescript with using objects, but that is fine

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
alex-ruskov-ft: I am not sure what the best threshold would be, but I would keep it quite low, like 0.1/0.2?

alex-ruskov-ft (approved):

PR #1163, opened 2025-04-28
feat: Emit exposure event from the Dropdown component

pdimova:
# Description
Jira:[ ENT-2379](https://financialtimes.atlassian.net/browse/ENT-2381)
This PR adds support for dispatching Amplitude exposure event for the Dropdown content. 

# Context
As part of an ongoing experiment, the exposure event tracking is implemented directly within the Dropdown component.

- [ ] I've discussed this feature with [the Platforms team](https://financialtimes.enterprise.slack.com/archives/C3TJ6KXEU)
- [ ] My branch has been rebased onto the latest commit on `main` (don't merge `main` into your branch)

# Notes
This is an experimental addition and may be refactored or removed based on the outcomes of the tracking initiative.

Screenshot 2025-04-28 at 14 45 13
Screenshot 2025-04-28 at 14 45 46

### DIFF
--- a/packages/dotcom-ui-header/README.md
+++ b/packages/dotcom-ui-header/README.md
@@ -77,8 +77,8 @@ All header components with the exception of `<LogoOnly />` require the following
 | showStickyHeader   | boolean                            | true     | true     | Enable rendering of the sticky header component                            |
 | showMegaNav        | boolean                            | true     | true     | Enable rendering of the drop-down "mega-nav"                            |
 | showAskButton      | boolean                            | true     | false    | Enable rendering of the ASK button                                             |
+| showProNavigation  | boolean - experimental             | true     | false    | Experimental Feature: Enable rendering of FT pro dropdown.                     |
 | data               | object                             | false    |          | Navigation data for rendering the header links fetched from the navigation API |
-| showProNavigation  | boolean | undefined - experimental | true     | undefined| Experimental Feature: Enable rendering of FT pro dropdown. Undefined means no experiment is conducted for a certain user cohort.
 | flags              | object                             | true     |          | Flags data passed from the server |
 
 
--- a/packages/dotcom-ui-header/README.md
+++ b/packages/dotcom-ui-header/README.md
@@ -71,14 +71,14 @@ All header components with the exception of `<LogoOnly />` require the following
 | variant            | 'simple' \| 'large-logo' \| string | true     | 'simple' | Adds a class name to the header element
 | userIsAnonymous    | boolean                            | true     | true     | Marks a user as anonymous - can be set by middleware included with n-express   |
 | userIsLoggedIn     | boolean                            | true     | false    | Marks a user as logged in - can be set by middleware included with n-express   |
-| userIsSubscribed     | boolean                            | true     | false    | Marks a user as subscribed - set by middleware(ammit task in preflight) included with n-express   |
+| userIsSubscribed   | boolean                            | true     | false    | Marks a user as subscribed - set by middleware(ammit task in preflight) included with n-express   |
 | showUserNavigation | boolean                            | true     | true     | Show user navigation options such as `Sign out` or `Subscribe`                 |
 | showSubNavigation  | boolean                            | true     | true     | Show the sub-navigation component which may include the crumbtrail             |
 | showStickyHeader   | boolean                            | true     | true     | Enable rendering of the sticky header component                            |
 | showMegaNav        | boolean                            | true     | true     | Enable rendering of the drop-down "mega-nav"                            |
 | showAskButton      | boolean                            | true     | false    | Enable rendering of the ASK button                                             |
 | data               | object                             | false    |          | Navigation data for rendering the header links fetched from the navigation API |
-| showProNavigation    | boolean - eperimental        |                   | true     | false    | Experimental Feature: Enable rendering of FT pro dropdown
+| showProNavigation  | boolean | 'control' | 'treatmen'  - eperimental              | true     | false    | Experimental Feature: Enable rendering of FT pro dropdown
 
 
 ## Header Elements
--- a/packages/dotcom-ui-header/README.md
+++ b/packages/dotcom-ui-header/README.md
@@ -78,7 +78,7 @@ All header components with the exception of `<LogoOnly />` require the following
 | showMegaNav        | boolean                            | true     | true     | Enable rendering of the drop-down "mega-nav"                            |
 | showAskButton      | boolean                            | true     | false    | Enable rendering of the ASK button                                             |
 | data               | object                             | false    |          | Navigation data for rendering the header links fetched from the navigation API |
-| showProNavigation  | boolean | 'control' | 'treatmen'  - eperimental              | true     | false    | Experimental Feature: Enable rendering of FT pro dropdown
+| showProNavigation  | boolean | undefined - experimental | true     | undefined| Experimental Feature: Enable rendering of FT pro dropdown. Undefined means no experiment is conducted for a certain user cohort.
 
 
 ## Header Elements
--- a/packages/dotcom-ui-header/README.md
+++ b/packages/dotcom-ui-header/README.md
@@ -79,7 +79,7 @@ All header components with the exception of `<LogoOnly />` require the following
 | showAskButton      | boolean                            | true     | false    | Enable rendering of the ASK button                                             |
 | showProNavigation  | boolean - experimental             | true     | false    | Experimental Feature: Enable rendering of FT pro dropdown.                     |
 | data               | object                             | false    |          | Navigation data for rendering the header links fetched from the navigation API |
-| flags              | object                             | true     |          | Flags data passed from the server |
+| metadata           | object                             | true     |          | Meta data passed from the Layer |
 
 
 
--- a/packages/dotcom-ui-header/README.md
+++ b/packages/dotcom-ui-header/README.md
@@ -79,6 +79,8 @@ All header components with the exception of `<LogoOnly />` require the following
 | showAskButton      | boolean                            | true     | false    | Enable rendering of the ASK button                                             |
 | data               | object                             | false    |          | Navigation data for rendering the header links fetched from the navigation API |
 | showProNavigation  | boolean | undefined - experimental | true     | undefined| Experimental Feature: Enable rendering of FT pro dropdown. Undefined means no experiment is conducted for a certain user cohort.
+| flags              | object                             | true     |          | Flags data passed from the server |
+
 
 
 ## Header Elements
--- a/packages/dotcom-ui-header/README.md
+++ b/packages/dotcom-ui-header/README.md
@@ -79,7 +79,7 @@ All header components with the exception of `<LogoOnly />` require the following
 | showAskButton      | boolean                            | true     | false    | Enable rendering of the ASK button                                             |
 | showProNavigation  | boolean - experimental             | true     | false    | Experimental Feature: Enable rendering of FT pro dropdown.                     |
 | data               | object                             | false    |          | Navigation data for rendering the header links fetched from the navigation API |
-| metadata           | object                             | true     |          | Meta data passed from the Layer |
+| metadata           | object                             | true     |          | Additional data context for side effects. Do not use for visible rendering effects |
 
 
 
--- a/packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
+++ b/packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
@@ -145,6 +145,38 @@ const trackDropdownView = (options) => {
 }
 
 /**
+ * Dispatches a custom event with tracking data for Amplitude experiments.
+ * 
+ */
+const trackDropdownExposure = () => {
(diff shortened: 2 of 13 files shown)


packages/dotcom-ui-header/README.md
```diff
@@ -71,14 +71,14 @@ All header components with the exception of `<LogoOnly />` require the following
...
 | showSubNavigation  | boolean                            | true     | true     | Show the sub-navigation component which may include the crumbtrail             |
 | showStickyHeader   | boolean                            | true     | true     | Enable rendering of the sticky header component                            |
 | showMegaNav        | boolean                            | true     | true     | Enable rendering of the drop-down "mega-nav"                            |
 | showAskButton      | boolean                            | true     | false    | Enable rendering of the ASK button                                             |
 | data               | object                             | false    |          | Navigation data for rendering the header links fetched from the navigation API |
-| showProNavigation    | boolean - eperimental        |                   | true     | false    | Experimental Feature: Enable rendering of FT pro dropdown
+| showProNavigation  | boolean | 'control' | 'treatmen'  - eperimental              | true     | false    | Experimental Feature: Enable rendering of FT pro dropdown
```
phelipemaia, line 81: **question:** it seems showProNavigation is not a boolean anymore? should we change to string?

also, why not keeping as boolean and leave the logic to whoever is consuming this? showProNavigation gives a sense of boolean not a flag value (control and treament).. we should probably stick with boolean here and the pages that use this pass the boolean according to the flag value they've loaded

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -144,6 +144,38 @@ const trackDropdownView = (options) => {
...
+const trackDropdownExposure = () => {
+  const flagProNavigation = document.querySelector('[data-flag-pro-navigation]')?.dataset.flagProNavigation
+
+  if (flagProNavigation !== undefined) {
+    const flagValue =
+      flagProNavigation === 'true'
+        ? 'treatment'
+        : flagProNavigation === 'false'
+        ? 'control'
+        : flagProNavigation
```
phelipemaia, line 160: **question:** what's the other possible value for flagProNavigation? could we simplify this and user one ternary only? if it's 'true' then it's 'treatment' all other cases are 'control'.

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -144,6 +144,38 @@ const trackDropdownView = (options) => {
   elementsToTrack.forEach((el) => observer.observe(el))
 }
 
+/**
+ * Dispatches a custom event with tracking data for Amplitude experiments.
+ * 
+ */
+const trackDropdownExposure = () => {
+  const flagProNavigation = document.querySelector('[data-flag-pro-navigation]')?.dataset.flagProNavigation
```
phelipemaia, line 152: **question:** correct if I'm wrong but data-flag-pro-navigation will only be in the DOM if the flag is variant and the page using page-kit has rendered the button? in that case, if the data-flag-pro-navigation is in the DOM you could just put the flagValue as 'treatment' straightaway...

phelipemaia (commented):
just a few things..

packages/dotcom-ui-header/README.md
pdimova: Yeah, this part is a bit tricky.
What we receive from next-flags is either false or the string 'treatment'. I’m not sure whether any other repositories are converting the 'treatment' string into a proper boolean.

In Workspace, we currently have the following check: 
`const showProNavigation = nextFlags['pro-navigation'] === 'treatment';`

This turns the flag into a boolean. However, since this is an Amplitude flag—which can have values beyond just booleans—it seems more appropriate to rename the variable to proNavigationFlag and pass it around in its original form. (this is different in nature from the on/off flags like proAsk etc.) 

Unfortunately, the first place where we encounter both value types (boolean and string) is in next-flags.

As I understand it, we need to send the strings 'control' and 'treatment' along with the exposure event.

My intention here is to retain whatever value comes from the consumer, and only transform the final value before sending it. This also gives us flexibility in case some consumers deviates from the ‘standard’ implementation.

It’s not a problem if each consumer performs the necessary checks and only passes a true or false—though that would mean each repository (or most of them) includes logic like `=== 'treatment'` somewhere in its code.

I’m happy to continue this discussion if needed

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
pdimova: The flag will be rendered in a hidden tag beneath the Dropdown / MyAccount button, depending on the flag’s value. The only time the flag will not be present in the DOM is when the user is not logged in.

phelipemaia (approved):

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -144,6 +144,33 @@ const trackDropdownView = (options) => {
...
 
+/**
+ * Dispatches a custom event with tracking data for Amplitude experiments.
+ *
+ */
+const trackDropdownExposure = () => {
+  const flagProNavigation = document.querySelector('[data-flag-pro-navigation]')?.dataset.flagProNavigation
+
+  if (flagProNavigation !== undefined) {
+    const flagExposureValue = flagProNavigation === 'true' ? 'treatment' : 'control'
```
olusoladavid, line 155: **question:** Can we use the `dotcom-ui-flags-embed` to retrieve this flag value?

pdimova: Thanks for pointing this out. I had a quick look at the  `dotcom-ui-data-embed` docs [here](https://github.com/Financial-Times/dotcom-page-kit/tree/main/packages/dotcom-ui-data-embed#client-side-integration)

Does this mean we need to add the following import in every repository where we use the Dropdown?
`import * as dataEmbed from '@financial-times/dotcom-ui-data-embed'`

packages/dotcom-ui-header/README.md
```diff
@@ -71,14 +71,14 @@ All header components with the exception of `<LogoOnly />` require the following
...
 | showSubNavigation  | boolean                            | true     | true     | Show the sub-navigation component which may include the crumbtrail             |
 | showStickyHeader   | boolean                            | true     | true     | Enable rendering of the sticky header component                            |
 | showMegaNav        | boolean                            | true     | true     | Enable rendering of the drop-down "mega-nav"                            |
 | showAskButton      | boolean                            | true     | false    | Enable rendering of the ASK button                                             |
 | data               | object                             | false    |          | Navigation data for rendering the header links fetched from the navigation API |
-| showProNavigation    | boolean - eperimental        |                   | true     | false    | Experimental Feature: Enable rendering of FT pro dropdown
+| showProNavigation  | boolean | undefined - experimental | true     | undefined| Experimental Feature: Enable rendering of FT pro dropdown. Undefined means no experiment is conducted for a certain user cohort.
```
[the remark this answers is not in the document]
olusoladavid: I see. Thank you. I think we can keep this prop as a simple binary switch as there will be no visual difference between `undefined (no experiment)` & `false (control variant)`. Also, setting the default to `false` means that if we remove the experimental feature tomorrow, it'd be as though it never existed. I'd prefer that we decouple a `show/hide` prop for this component from the experimentation values. This [blog post ](https://martinfowler.com/articles/feature-toggles.html#ImplementationTechniques) does a decent job of explaining the usefulness of such decoupling.

We can extract the experimentation values using the embedded flags JSON as I explained in another comment

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
pdimova: As agreed, `flagsData` property will be added to the Layer component. Please check out the updated implementation

phelipemaia (approved):

phelipemaia (approved):

alex-ruskov-ft (approved):

packages/dotcom-ui-layout/README.md
```diff
@@ -86,6 +86,7 @@ This component includes styles written in Sass which includes the styles the [he
 | footerOptions   | TFooterProps                                    | true     | `undefined` | Pass options to the footer component                                                         |
 | footerComponent | ReactElement                                    | true     | `undefined` | Pass a custom footer                                                                         |
 | contents        | string                                          | true     | `undefined` | A prerendered string of HTML used to insert the page contents when not using JSX composition |
+| metadata        | object                                          | true     | `undefined` | Meta data passed from the server |
```
olusoladavid, line 89:
```suggestion
| metadata        | object                                          | true     | `undefined` | Additional data context for side effects. Do not use for visible rendering effects |
```

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -144,6 +144,57 @@ const trackDropdownView = (options) => {
   elementsToTrack.forEach((el) => observer.observe(el))
 }
 
+/**
+ * Dispatches a custom event with tracking data for Amplitude experiments.
```
olusoladavid, line 148:
```suggestion
 * Dispatches a custom exposure event for Amplitude experiment.
```

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -144,6 +144,57 @@ const trackDropdownView = (options) => {
...
+ */
+const trackDropdownExposure = () => {
+  const flagDataProNavigation =
+    document.querySelector('[data-flag-pro-navigation]')?.dataset.flagProNavigation ||
+    (document.querySelector('#page-kit-app-context') &&
+      JSON.parse(document.querySelector('#page-kit-app-context').innerText).abTestState?.match(
+        /pro-navigation:([^,]+)/
+      )?.[1])
+
+  if (flagDataProNavigation !== undefined && flagDataProNavigation !== 'no-experiment') {
```
olusoladavid, line 160:
```suggestion
  if (flagDataProNavigation && flagDataProNavigation !== 'no-experiment') {
```
Also, curiously, what is the `no-experiment` value? Is it a variant?

packages/dotcom-ui-header/src/components/top/partials.tsx
```diff
@@ -123,6 +123,7 @@ const TopColumnRightLoggedIn = (props: THeaderProps) => {
       {signInAction && (
         <MenuButton
           showProNavigation={props.showProNavigation}
+          flagMetadataProNavigation={props.metadata?.['pro-navigation']}
```
olusoladavid, line 126: **suggestion:** should `flagMetadataProNavigation` just be named as `proNavigationVariant` ?

olusoladavid (commented):
Great work @pdimova ! Lgtm. Had some questions/comments but overall looking good

packages/dotcom-ui-header/src/components/top/partials.tsx
pdimova: Done

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
```diff
@@ -144,6 +144,57 @@ const trackDropdownView = (options) => {
...
 
+/**
+ * Dispatches a custom event with tracking data for Amplitude experiments.
+ *
+ * Remove all relevant code when the experiment is complete.
+ */
+const trackDropdownExposure = () => {
+  const flagDataProNavigation =
+    document.querySelector('[data-flag-pro-navigation]')?.dataset.flagProNavigation ||
+    (document.querySelector('#page-kit-app-context') &&
```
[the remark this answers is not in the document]
pdimova: I believe we agreed to obtain the `pro-navigation` flag value from the `abTestState` string in the context for all `next-*` pages. This way, we can keep the amount of code added to the repos as minimal as possible. Unfortunately, we need to keep the metadata object so that the professional pages can access the value on the client side as well since they do not use the context.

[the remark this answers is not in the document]
olusoladavid: No problem. I was hoping we'd pass the value in as a prop from the `next-*` pages but happy to keep as-is, given that this code will be short-lived and scoped to the dropdown. Thank you

packages/dotcom-ui-header/src/components/dropdown-navigation/dropdownNavigation.js
olusoladavid: This code will be removed after the experiment right? Since there will be no need for exposure event afterwards?

pdimova: That's the plan. 
The pro-navigation flag variable (proNavigationVariant) will be removed; only the dropdown toggler prop (showProNavigation) will remain. 
No further next-* PRs should be necessary, ideally.

olusoladavid (approved):
🚢

PR #1173, opened 2025-09-03
CULT-477: change error type on get editions failure

danieleruiz:
### Description

**Problem**

We see errors matching this description in out apps
Error: The provided edition "asxxp1" is not a valid edition
These usually comes from bots, automation and malicious traffic
And results in 500 response from our servers
 It ought to be classes as a 400 Bad Request client error

Impact
In our metrics and logs it makes it looks like there’s a problem that needs to be actioned.
Investigating problems that are actually client errors is a waste of time.
And causes noise when there are real problems to diagnose

### Ticket
https://financialtimes.atlassian.net/browse/CULT-477


danieleruiz:
> @dotcom-reliability-kit/errors

replaced @kavanagh

PR #1175, opened 2025-09-09
CPP-2651 update skip link text to "skip to main content"

apaleslimghost:


olusoladavid (commented):
Looks like we are missing one more integration test update - https://github.com/Financial-Times/dotcom-page-kit/blob/5cc4e906af0ab648e7e4c00fcea3f44cfb3d3f0e/examples/ft-ui/__test__/integration.test.js#L24

olusoladavid (approved):

PR #1176, opened 2025-09-16
CPP-2652 set tabindex="-1" on sticky header elements missing it

apaleslimghost:
two unrelated issues with the same symptoms:

- the Ask FT button didn't have a `sticky` variant and wasn't setting `tabindex` at all
- the sign in button did have a `sticky` variant, but the anon header partial wasn't propagating the variant to the component

### DIFF
--- a/packages/dotcom-ui-header/src/__test__/components/DropdownNavigation.spec.tsx
+++ b/packages/dotcom-ui-header/src/__test__/components/DropdownNavigation.spec.tsx
@@ -15,6 +15,7 @@ describe('DropdownNavigation', () => {
         options={PRO_NAVIGATION_DROPDOWN_DEFAULT_LIST}
         headerContent="Test Title"
         trackingKey="test_tracking"
+        variant='default'
       />
     )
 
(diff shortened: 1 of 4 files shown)


olusoladavid (approved):
Lovely. Thank you. In the FT Pro experience, the Pro DropDownNavigation button still explicitly (tabindex=0) traps focus in the sticky header. I guess we can leave this for the Pro team to sort out. You can preview by activating `proNavigation` experiment flag in Toggler.

Would be interesting to discuss how we ensure that people adding interactive elements in the header don't forget to remove their elements from the natural tabbing order in the future

apaleslimghost:
good point! i might as well fix the dropdown in this PR. i'll also add a test to check that every interactive element in the sticky header has `tabindex=-1`

olusoladavid:
> this is proving to be very tricky, we're still using react-test-renderer for the tests and so we'd have to manually walk the tree (and i can't get it working). i don't think it's worth refactoring to use @testing-library/react just for this test.

Thanks @apaleslimghost - yeah happy to leave it as-is for now.. we can probably add a small reminder to the pull request template but not worth additional dev investment as you have mentioned

